PR #10, opened 2014-12-09
Show JID as contact name if instead an empty string would have been show...

rarog:
...n.


wrooot:
I have filed SPARK-1607 for this. Will try to find a way to commit your changes to the master.

wrooot:
rarog, i'm trying to understand what scenario your change is fixing. Can you comment here or on the ticket https://igniterealtime.org/issues/browse/SPARK-1607 what exactly is happening now and what your patch will do?

rarog:
Done, I hope this is enough information and bug description. :)

PR #29, opened 2015-08-14
Add TranscriptWindowTextProcesser

bonashen:
Add TranscriptWindowTextProcesser,let's develope plugin process TranscriptWindow text.


wrooot:
Hi. Sorry for late reply (stuck in the hospital..). I'm not a developer, so can you explain in general what you patch adds to Spark?  How will it back nefit other users or is more for you personal needs? Will it affect current Spark behavior? Also is it a normal practice to use .* wildcards in imports? 

I will have to file a ticket in the bug tracker for your patch after you explain it.

wrooot:
Anything to reply?

marclaporte:
@bonashen  We really want you to get involved more!

wrooot:
Closing as no further feedback received and it already conflicts with the current code.

PR #34, opened 2015-09-14
Fix chat messages not being shown when there are no old history showing

wrooot:
https://igniterealtime.org/issues/browse/SPARK-1631


wrooot:
https://igniterealtime.org/issues/browse/SPARK-1631

PR #40, opened 2015-10-02
IzPack 5 Support

veganaize:
I've added "installer.izpack5" to the build/build.xml file.

I've also created an XML file for the new IzPack 5 syntax.


wrooot:
I've merged your pull request. Filed as https://igniterealtime.org/issues/browse/SPARK-1633
Thanks for the patch.

PR #56, opened 2015-11-25
Add scrollbar for Emoticon pane

idma88:


wrooot:
My only nitpick is the introduced frame around the emoticons picker. Doesn't look pretty. Maybe it can be removed (at least when scrollbar is not needed/shown)?

wrooot:
Filed as https://igniterealtime.org/issues/browse/SPARK-1652

wrooot:
Also, adding:

scrollPane.getViewport().setBackground(Color.WHITE);

after: JScrollPane scrollPane = new JScrollPane(gridContainer);

fixes the gray background problem ;)

wrooot:
Solution for my border nitpick.

import javax.swing.BorderFactory;

and add

scrollPane.setBorder(BorderFactory.createEmptyBorder());

PR #60, opened 2015-12-14
New translations and minor changes

idma88:


wrooot:
Merged. I know that translations are hard thing to finalize and can be improved all the time. But next time add a comment, that this is still work in progress or accumulate all translations in one commit if possible, so i don't have to reopen or file many tickets for this ;)

PR #67, opened 2015-12-22
Another try at updating smack to 3.4.1

speedy01:


wrooot:
Apparently Smack's internal version hasn't been changed for 3.4.1. I have compared sizes of what is now included in Spark latest build and in smack 3.4.1 zip and they are the same. So we will consider this an update to 3.4.1. Only it will shows 3.4.0 in the About window.

PR #69, opened 2015-12-29
SPARK-1091 mac dmg builder

speedy01:
commented out since it wasn't working with Bamboo remote agent.  Same temp fix appears to have been used with the openfire dmg builder. Unable to test.


wrooot:
Well, i still don't know where in Bamboo dmg should be produced, but lets hope it will make dmg installer possible again.

PR #92, opened 2016-02-21
Spark-1701, Spark-1696, Spark-1620

speedy01:
This should resolve the plugin/sound/emoticon issues with the mac build to closer mimic the same behavior as the other platforms.  I tested this on windows and max os x, but did not test it with linux.


wrooot:
Looks fine on linux too.

PR #106, opened 2016-03-13
SPARK-1714 add NPE check

Alexander198961:
Added NPE check


wrooot:
This PR is using changes made by your previous PR, which i have reverted, so i would have to use cmd to fix this. I will close this PR and apply your changes via my own PR (would be faster than installing Git cmd and doing all the magic). I have already tested this and it works fine this time. Thanks for the patch. Hope you will find a way to go around Fastpath issue in your previous patch for broadcast ;)

PR #108, opened 2016-03-14
Some fixes design and translation

idma88:


wrooot:
Going to close this PR. I will add fix for Business card design as a separate commit - https://github.com/igniterealtime/Spark/pull/112

I don't agree on a second change. SparkToaster is the name of a type of popup (original Spark's popup, called Toast). Yeah, it can probably be called some other way and even be translatable, but certainly not "Top right" :)

In future do not add different changes into one PR, do separate PRs. Also, try to have only ne commit per PR or at least squash them into one commit before doing PR.

PR #111, opened 2016-03-17
Fix for SPARK-1637

Alexander198961:
Seems tab switched to red color now https://igniterealtime.org/issues/browse/SPARK-1637

### DIFF
--- a/src/java/org/jivesoftware/sparkimpl/plugin/alerts/BroadcastPlugin.java
+++ b/src/java/org/jivesoftware/sparkimpl/plugin/alerts/BroadcastPlugin.java
@@ -319,6 +319,7 @@ public class BroadcastPlugin extends SparkTabHandler implements Plugin, PacketLi
     private void userToUserBroadcast(Message message, Type type, String from) {
 	String jid = StringUtils.parseBareAddress(from);
 	String nickname = SparkManager.getUserManager().getUserNicknameFromJID(jid);
+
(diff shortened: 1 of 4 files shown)


wrooot:
There are a few problems. First one is Git related. I have reverted your last PR for broadcast issue because of build failing for Fastpath. This PR is based on those changes, which are not in the source. So it conflicts (same as was with the PrivateNotes patch). Not sure what can be done to avoid that (maybe syncing or cloning a fork a new and resubmitting whole patch with updates). I'm not a Git expert.

So, to avoid problems i have manually applied both old PR and this one to my local branch in Eclipse and tested it. Plugins build fine now. It also fixes the "receiving broadcast into existing tab doesn't change tab title into Broadcast from..". But the existing tab still doesn't flash. It also makes SPARK-1686 to happen more often (it just shows Broadcast from for second and switches to a simple Name tab immediately, when receiving a broadcast into new tab). It doesn't fix SPARK-1636 (although it probably wasn't meant to fix that, just mentioning that), receiving broadcast into a first tab still doesn't append "Broadcast from..". And it introduces new problem. Now when you receive a simple message a tab stays red until you reply.

I will close this PR. But i will save the changes in my local branch and wait, maybe you will come up with a fix for this. You can post new changes as a PR (it will conflict again ,so i would just copy the changes to my branch) or as a comment on the ticket or here. If not, i will have to discard all the changes eventually.

PR #114, opened 2016-03-23
SPARK-1563 Continued to import additional fixes from cstux patch

speedy01:
Imported a few more commits from cstux patch.  This does not address jingle, however it appears to correct some issues with the sip phone.


wrooot:
Guus and CSH say that doing more than one commit per PR is bad. They say you can squash them into one commit (i have tried in Eclipse, but it didn't work for me). Will merge this anyway.

wrooot:
Also, RPM spec file should be updated when adding or removing libraries.

PR #116, opened 2016-03-25
Duplicate conditional in NetworkAddressManager.java

kilida:
Looks like this should be `==` when returning 0.


wrooot:
@speedy01 this looks like a good typo catch, but as it is related to SIP, i want your approval :)

speedy01:
looks good :)

PR #121, opened 2016-03-30
Update of German translation

rarog:
Many strings translated, corrected or otherwise fixed. More to come.


wrooot:
Merged. Next time, please, do a single commit or squash commits.

PR #130, opened 2016-04-25
Roar updates

wolfposd:
SPARK-1596 different settings for groupchats
SPARK-1599 keyword based matching


wrooot:
Heh, i saw you have forked Spark recently, so thought maybe you will do something and here it is :) Nice to see you coding for Spark again, Wolf. Will apply and test this PR later in the day ;)

wrooot:
Keyword popup settings apply only when single-user chat popups are disabled. This covers second part of SPARK-1599, but i think some users might want to still have regular popups and then keyword popups with different color settings. So, i think keyword settings should apply even if single-user popups are enabled.

Also roar.group says Mutli-user. Should be Multi-user :)

wolfposd:
will update

PR #131, opened 2016-04-26
Roar updates

wolfposd:
- reordered the popup-detection (1.keyword, 2. single, 3. MUC)
- added OS-dependent notifications (looks nice on Win10 and OSX10.8+)
- refactored code


wrooot:
Merged. A few quirks. When System-notification is selected, the explanation message is so wide that horizontal scroll bar appears. Maybe it can be shorter or in two lines. Also different styles apply only after Spark restart. Maybe it should say that somewhere (or maybe even force a restart like when skin is changed from substance to jtattoo or when language is changed). That's minor things. I'm going to close tickets now. But you can still update it, if you wish.

wrooot:
SPARK-1596
SPARK-1599

wrooot:
Hey, Wolf. Just wondered if you could be able to take a look at this https://issues.igniterealtime.org/browse/SPARK-1755 We have updated Spark to Smack 4 and had to fix Roar because messages not showing up on fresh profile. And one of these changes introduced that redundant popups issue.

wrooot:
Actually, Guus has just fixed that. But there are other tickets :) https://issues.igniterealtime.org/browse/SPARK-1782 and https://issues.igniterealtime.org/browse/SPARK-1784

PR #133, opened 2016-04-29
Spark 1735 small UI improvements

guusdk:
A number of fixes for MUC, mostly.


wrooot:
What about line ending? I do copy paste in Eclipse and it doesn't show that something is wrong with line endings.

guusdk:
Probably a your-Windows-vs.-my-Linux problem. Github keeps insisting that files have changed, even when I revert the changes. Oh well. :)

PR #134, opened 2016-04-29
SPARK-1613: Force wrapping of long lines.

guusdk:


wrooot:
You are our hero! :+1: Although, i would prefer for it to cut on the border of the window instead of going to the next line for the whole long line, but as it would most often happen with URLs, it's a minor thing and is for sure much much better than a vertical scroll :)

guusdk:
Yeah - I'm not a real expert in these UI things. Best effort, couldn't get it prettier :)

nokaoii:
Thank you! You're awesome guusdk!

wrooot:
There is a little regression uncovered by speedy. The messages don't come through on a fresh profile.

PR #135, opened 2016-05-03
SMACK4: Prevent NPE in Roar plugin.

guusdk:
This commit fixes the problem for which SPARK-1613 was re-opened (the actual problem does not relate to SPARK-1613 though).


guusdk:
For posterity: @wrooot created a separate issue for the problem: SPARK-1736.

guusdk:
Also, unlike the commit message, it has nothing to do with Smack 4.

PR #136, opened 2016-05-04
Smack 4

guusdk:
This collection of commits migrates the Smack library that is used by Spark from version 3 to 4.

These commits are primarily a one-on-one translation of functionality. There's room for improvement here, as Smack 4 brings many new features that are not used by Spark. I suggest we take that out of the scope of this PR though.

These commits bring the Spark code base in a state where it compiles, and passes some basic functionality tests (I have been using it for a couple of days) - but there will be problems. I suggest we merge nonetheless, as that will make the changes more accessible to the general development public.


wrooot:
Crossing fingers :)

PR #137, opened 2016-05-05
SPARK-1739

speedy01:
@guusdk  Please review..In my test, this restores sparks previous behavior of allowing self signed certs, etc over tls.

### DIFF
--- a/src/java/org/jivesoftware/LoginDialog.java
+++ b/src/java/org/jivesoftware/LoginDialog.java
@@ -212,46 +212,46 @@ public class LoginDialog {
 
         ProxyInfo proxyInfo = null;
         if (localPref.isProxyEnabled()) {
-        	ProxyInfo.ProxyType pType = localPref.getProtocol().equals("SOCKS") ?
-        		ProxyInfo.ProxyType.SOCKS5 : ProxyInfo.ProxyType.HTTP;
-        	String pHost = ModelUtil.hasLength(localPref.getHost()) ?
-        		localPref.getHost() : null;
-        	int pPort = ModelUtil.hasLength(localPref.getPort()) ?
-        		Integer.parseInt(localPref.getPort()) : 0;
-        	String pUser = ModelUtil.hasLength(localPref.getProxyUsername()) ?
-        		localPref.getProxyUsername() : null;
-        	String pPass = ModelUtil.hasLength(localPref.getProxyPassword()) ?
-        		localPref.getProxyPassword() : null;
-        	
-        	if (pHost != null && pPort != 0) {
-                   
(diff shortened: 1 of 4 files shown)


src/java/org/jivesoftware/LoginDialog.java
```diff
@@ -277,11 +239,22 @@ protected XMPPTCPConnectionConfiguration retrieveConnectionConfiguration() {
         }
 
         final XMPPTCPConnectionConfiguration.Builder builder = XMPPTCPConnectionConfiguration.builder()
-                .setUsernameAndPassword( "username", "password" )
+                .setUsernameAndPassword( loginUsername, loginPassword )
```
guusdk: Wow. I wonder why it is working for me in the first place (no, my username is not "username").

src/java/org/jivesoftware/LoginDialog.java
```diff
@@ -277,11 +239,22 @@ protected XMPPTCPConnectionConfiguration retrieveConnectionConfiguration() {
...
                 .setServiceName( loginServer )
                 .setPort( port )
                 .setSendPresence( false )
                 .setCompressionEnabled( localPref.isCompressionEnabled() );
+        try {
+            TLSUtils.acceptAllCertificates(builder);
```
guusdk: Lets make this configurable. Please add an `isAcceptAllCertificates()` and `setAcceptAllCertificates(boolean accept)` and make it default to 'true'. You can then add an "if" condition to this line.

For brownie points, add the setting to `LoginSettingDialog`. :)

src/java/org/jivesoftware/LoginDialog.java
```diff
@@ -277,11 +239,22 @@ protected XMPPTCPConnectionConfiguration retrieveConnectionConfiguration() {
...
+        } catch (NoSuchAlgorithmException e) {
+            e.printStackTrace();
```
guusdk: Instead of `e.printStackTrace();`  please do `Log.warn( "Unable to create configuration.", e );` which will make sure that the error ends up in the logfiles (As a rule of thumb, never do `e.printStackTrace()` ).

speedy01:
@guusdk  I think I implemented your suggestions.  I tested locally, and it seemed to work.

guusdk:
Looks good to me. I'll leave it for @wrooot to have the final say.

wrooot:
This PR is conflicting currently (probably after i have merged Guus code refactoring (imports, etc.)).

speedy01:
i have no idea who to resolve this conflicts!

guusdk:
I think I fixed this in https://github.com/igniterealtime/Spark/pull/146

PR #138, opened 2016-05-06
SPARK-1741: Improve on code structure

guusdk:
These commits simplify the code of Spark by making use of more modern code constructs that are now available in Java. No functional changes should be introduced by these changes (with the exception of one potential bugfix that relates to string comparison).


wrooot:
Bamboo fails to build Install4J, Mac installers. http://bamboo.igniterealtime.org/browse/SPARK-INSTALL4J-818/log

guusdk:
I noticed that. That one still depends on Java 7 - my changes require Java
8. I'll try to update the Bamboo agents.

PR #161, opened 2016-05-18
Spark 1729

Alexander198961:


guusdk:
Hello - thanks for your contribution. There are a number of things to address:

- There is a conflict with the master branch. Could you please rebase your commits?
- You appear to introduce both java.util.logging as well as Jive's logger. One will do. :)
- When logging an exception, please make sure that the stack trace is also logged!
- There is quite a bit of whitespace commits. Perhaps you can avoid those?

Finally, as your commit messages do not appear to be very descriptive, you might want to squash all commits into one.

Alexander198961:
hello fixed onflict with the master branch

guusdk:
Thanks, that was good. Please, address the other issues too. Also, in your last change, you added a couple of new files that should not be included at all. Please remove those.

Alexander198961:
removed unnecessary files

wrooot:
Why do you remove a shortcut from Save password checkbox (&) ? I think it should stay there.

wrooot:
Applied this patch manually as https://github.com/igniterealtime/Spark/pull/163 (removed unneeded logger, fixed whitespace commits, fixed Save password checkbox, added comments). Though i think this can be still improved as it currently opens and reads preferences file twice to get the room password and there is no way to remove the saved password other than deleting it from the preferences file.

PR #166, opened 2016-05-29
Italian translation

nicoben:
Now I created a dedicated branch for translation. I shouldn't mess up your work anymore. 
Thank you for your patience ;)


wrooot:
Yeah, now it's better. No conflicts ;)

PR #180, opened 2016-07-25
Spark 1769

speedy01:
updated substance (insubstantial) to 7.3 using binaries from maven.org/maven2/com/github/insubstantial


wrooot:
Are you sure this is only Insubstantial update? I see a lot of changes. Or are they needed for new LaF to work?

speedy01:
In my testing, I had to include the extra libraries.  Also, the changes to the few files are small (but due to my IDE, it looks worse than it is!). They were made as a result of some bugs that that popped up when I tested it.  I enclosed some of the ui elements in EventQueue.invokeLater(() -> { } to resolve them.
check ConferenceServices.java around line 162 to see what I mean.

wrooot:
I saw that the changes were small and looked like UI timing improvement. It is just, that it is easier to track changes, when they are related to the ticket. I guess we can call it a part of LaF update :)

IvanRF:
Hi guys, by chance I ended up reading some of the threads in your [community](https://community.igniterealtime.org/search.jspa?q=substance) about Substance.
How do you feel about taking control over the [Insubstantial project](https://github.com/Insubstantial/insubstantial)? I ask because your project is active and you have several contributors :smiley:

wrooot:
We don't have enough manpower (or developers) to move forward our projects. I don't think we can take control of Insubstantial. Btw, if you want to continue this discussion, it is better to do this in our forums (General) or group chat, so we don't clutter PR comments with a side topic :wink:

IvanRF:
@wrooot ok, it's a pity that project remains unmaintained. This was the fastest way for me to contact you. Regards!

PR #213, opened 2016-09-01
Italian translation

nicoben:
More italian translations


wrooot:
Once again, you PR conflicts with the current source, so i have manually applied it as https://github.com/igniterealtime/Spark/pull/214 Maybe you have started this some time ago and there where changes to translations files recently, so you used older state for your PR.

PR #227, opened 2016-10-02
SPARK-1802: Add littleluck LaF skin

freeseawind:
Add littleluck LaF skin


wrooot:
Thanks for your updated patch. Will wait for @guusdk to review it.

freeseawind:
Thanks for the response.

src/littleluck/src/java/freeseawind/swing/package.html
```diff
@@ -0,0 +1,7 @@
+<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN">
+<html>
+<head></head>
+<body>
+本包内存放的是Swing的扩展实现类。
```
guusdk: Please replace this with an English text.

src/littleluck/src/java/freeseawind/swing/LuckScrollPaneLayout.java
```diff
@@ -0,0 +1,101 @@
...
+import java.awt.Container;
+import java.awt.Insets;
+import java.awt.Rectangle;
+
+import javax.swing.JLayeredPane;
+import javax.swing.JScrollPane;
+import javax.swing.ScrollPaneLayout;
+
+/**
+ * ScrollPane布局实现类，主要针对{@link LuckScrollPane}实现
```
guusdk: Please replace this with English text.

guusdk:
Additionally, it would be good to squash all of those commits in a few commits that are more 'atomic' in nature.

wrooot:
I can squash it when merging, if that helps.

guusdk:
Lets do that after all other changes.

freeseawind:
I will remove Chinese characters in source code,  is anything else still needs to be improved?

guusdk:
Please make all code that is not specific to Spark a JAR file. I would like
to include your code as a library, not as "part of" Spark. Only the code
that is specific to Spark should not be in that JAR file.

freeseawind:
Your mean I only commit `laf.jar `and `ThemPanel`. I am not very clear which part is is specific to Spark.I think commit source code then everyone can read and update it is nice way.

wrooot:
To be fair, i think freeseawind is taking Jtattoo as an example. And Jtattoo has been added with a bunch of java files instead of one JAR (as Subtance skin has been added - see at build\lib\dist\substance.jar).

I think Guus wants you to add your skin functionality (images, java files) with a single JAR file and only add changes like ThemePanel.java, license, build (not needed probably also) as the direct changes to Spark's source).

Yeah, if committing your source code directly, then we can modify that source, but that's not the best practice. You are the maintainer of that project/skin, so the changes has to come from you as a new version of JAR. This is easier to update than importing all the changes JAVA files, images, etc.

guusdk:
Exactly. If you want other developers to work on the LittleLuck code (which is a good idea!), you should create a new Github project for just that code (or on another place like Github). That way, other projects can also use LittleLuck.

freeseawind:
Ok, I know what I can do. I will recommit.Do I need to close this submission first?

guusdk:
Whatever you prefer.

guusdk:
Awesome, that cleans up a lot. Now, the total number of files changed is two. :)

Could you squash all of your commits into one? There is no point in having 17 commits that are all undone by the last one. :)

Also, it would be cool to have a link to your project page. This way, people that are interested in your code can find it easily. Perhaps you can add a comment with a link to your Github-page somewhere?

PR #235, opened 2016-10-23
SPARK-1021 Default to Name field as nickname when adding a contact

wrooot:


raspoutine1984:
Hi,
This patch have solved my issue when the user A want to add the user B but when the user B receive the request the username appear in nickname field, instead of Name.
Is it possible to add this feature?

Thanks a lot

wrooot:
I might look into it, but can't promise anything. This patch was easy as i've been pointed to the exact code to change it. Not sure i will be able to find where the request form is triggered.

raspoutine1984:
Thanks for your reply. 
I have searched in the sources and I think the solution is here : https://github.com/igniterealtime/Spark/blob/master/src/java/org/jivesoftware/spark/ui/SubscriptionDialog.java
But next, it's not in my skills :-|

wrooot:
I have filed https://issues.igniterealtime.org/browse/SPARK-1842 but i don't know how to do it.

raspoutine1984:
Hello,

OK, many thanks I will monitor this ticket.

raspoutine1984:
Hello wrooot,

Any news concerning this request? I'm ready to deploy Openfire and Spark on my network, it's the last thing who block us. I think that would be benefic for many other users? 

Regards

wrooot:
We don't have experienced full time java developers working on Spark. I don't know how to achieve that. So, you might be waiting a few years before someone submits a patch for this.

PR #238, opened 2016-10-26
SPARK-1520

speedy01:
This should get the presences and return system icon to the correct icon.


src/java/org/jivesoftware/sparkimpl/plugin/systray/SysTrayPlugin.java
```diff
@@ -58,6 +59,7 @@
 import org.jivesoftware.sparkimpl.settings.local.SettingsManager;
 import org.jivesoftware.smack.chat.ChatManagerListener;
 import org.jivesoftware.smackx.chatstates.ChatStateListener;
+import org.lobobrowser.html.domimpl.HTMLElementBuilder;
```
akrherz: What's going on with this new import?  Sorry for the ignorance

wrooot: yeah, it doesn't look that it is needed there

speedy01: Yeah, that import should not be there . Im not sure how that got in there ! Ill remove it!

wrooot:
speedy, sorry, i have botched your PR with a revert i had to do to another PR. Could you please submit new one?

PR #239, opened 2016-10-27
Fix SPARK-1051

Alexander198961:
https://issues.igniterealtime.org/projects/SPARK/issues/SPARK-1051


wrooot:
This is not what was requested. Your change sends the same broadcast to yourself. The request was about being able to check whom you have sent the request to. So it has to be saved into chat history of a user you are sending broadcast to (but on your side). Sending broadcast to yourself doesn't provide such information. Moreover, you don't usually have yourself in your roster, so after closing the chat window you won't be able to check the history. And just in general, getting your own message is just bizarre :) Reverting this change.

PR #243, opened 2016-10-28
Spark 1051

Alexander198961:
The message from sender side now looks like " 
(01:30) test1@www.freevoip.od.ua - Broadcast: test2@www.freevoip.od.ua test4@www.freevoip.od.ua :Hello broadcast"  But still maybe need some color modification

### DIFF
--- a/src/java/org/jivesoftware/sparkimpl/plugin/alerts/BroadcastDialog.java
+++ b/src/java/org/jivesoftware/sparkimpl/plugin/alerts/BroadcastDialog.java
@@ -59,7 +59,6 @@ import org.jivesoftware.spark.ui.ContactItem;
 import org.jivesoftware.spark.ui.ContactList;
 import org.jivesoftware.spark.util.ModelUtil;
 import org.jivesoftware.spark.util.log.Log;
-
 /**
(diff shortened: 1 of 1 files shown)


wrooot:
Alexander, i understand what you are trying to achieve, but i don;t think this is what SPARK-1051 is asking for. And again, having a message popping up for you as a new message, when you have actually sent it to others just looks messy. And most importantly, when you close this window, there is no way to open it again as user don't usually have themselves on their roster. This lead me to another idea to solve this kind of. I have filed https://issues.igniterealtime.org/browse/SPARK-1825 which could work as a workaround maybe.

wrooot:
Alexander, is Fix SPARK-1803 commit enough to fix logins with non-latin usernames? Or does it depend on https://github.com/igniterealtime/Openfire/pull/657 as well?

Alexander198961:
don't catch you . in short 
the pull https://github.com/igniterealtime/Openfire/pull/657 has solved problem with login with non-latin usernames there was encoding issue .

wrooot:
"don't catch you" - это значит что не понимаешь? Могу по-русски объяснить (на счет 1051).

PR #251, opened 2016-11-01
SPARK-1516 - Spark should not let open profile edit by clicking the avatar if profile changing is disabled

BittyByte:
Added check for enterprise client control feature VCARD_FEATURE so that clicking on avatar will not allow profile editing.

### DIFF
--- a/src/java/org/jivesoftware/resource/default.properties
+++ b/src/java/org/jivesoftware/resource/default.properties
@@ -109,6 +109,27 @@ DISABLE_AVATAR_TAB = false
 # If true, disable the "Plugins" menu item
 DISABLE_PLUGINS_MENU_ITEM = false
 
+# If true, disable the option to transfer files
+DISABLE_FILE_XFER = false
+
+# If true, disable the option to remove contacts and groups in roster
+DISABLE_REMOVALS = false
+
+# If true, disable the option to rename contacts and groups in roster
+DISABLE_RENAMES = false
+
+# If true, disable the options to move and copy contacts
+DISABLE_MOVE_AND_COPY = false
+
+# If true, disable the option to edit the profile
+DISABLE_EDIT_PROFILE = false
+
+# If true, disable the "View notes" option under "Actions"
+DISABLE_VIEW_NOTES = false
+
+# If true, disable the "View task list" option under "Actions"
+DISABLE_VIEW_TASK_LIST = false
+
 #################################################
 ################## File Transfer ################
 #################################################
--- a/src/java/org/jivesoftware/resource/default.properties
+++ b/src/java/org/jivesoftware/resource/default.properties
@@ -112,6 +112,9 @@ DISABLE_PLUGINS_MENU_ITEM = false
 # If true, disable the option to transfer files
 DISABLE_FILE_XFER = false
 
+# If true, disable the option to take screenshots
+DISABLE_SCREENSHOTS = false
+
 # If true, disable the option to remove contacts and groups in roster
 DISABLE_REMOVALS = false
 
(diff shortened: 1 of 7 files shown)


wrooot:
Well, as in my tries, this doesn't work. Avatar is still clickable when i disable profile\avatar change in Client Control (though Edit profile menu disappears as it should). In VCardmanager.java there is some sort of initialization, and there is none in StatusBar.java. Maybe it is needed, as it seems that for StatusBar.java VCARD_FEATURE value is always "true".

BittyByte:
Hi Wroot, I will test this thoroughly and try to resolve the issue as soon 
as possible.

Sent with AquaMail for Android
http://www.aqua-mail.com

wrooot:
No pressure ;) I just thought it would be easy to fix that little issue along with your other patches. But it seems it is not as easy.

BittyByte:
I'm going to have a look at the server-side plug-in to see if perhaps the 
problem is actually there. When I made the mod, I wasn't connected to an 
openfire server so I wasn't able to test it and just assumed it should 
work. Sorry about that...

Sent with AquaMail for Android
http://www.aqua-mail.com

wrooot:
As i've said the menu disappears, so the server is sending something actually. But the statusbar part is not getting it.

BittyByte:
In the course of trying to resolve this issue, I came across yet another bug which occurs when disabling broadcasts in the client control plugin. It seems that when broadcasts are disabled, the user can still highlight multiple contacts and right-click to select the "Send broadcast to select users" menu option. I have fixed that issue, but I'm still stumped as to why all the client control feature variables in StatusBar.java are always set to 'true'. 

Anyway, it's after 3:00AM here so I'll continue to debug this later today... <yawn>

wrooot:
That's ok. We can leave the check you have added to StatusBar.java and maybe someone else will figure it out later (it won't harm leaving it). As about broadcasting, the option in Client Control states: "Allow for broadcasting of messages to specified groups or entire roster." so it doesn't imply that sending broadcasts to selected users should also be disabled. At least that's why i haven't filed a ticket for that in the past :) Though it is probably logical to not allow that either. We can block that also and then i can also submit a PR to fix wording in Client Control plugin. You can file a ticket for this in JIRA on your own (or i can file it if you wish).

BittyByte:
Hi Wroot,

 

I enjoy a good mystery! It’s very strange how all 3 of those Client Control feature values are always set to ‘true’ in the StatusBar.java file, while they show the correct values in other files. In an attempt to understand the code better, I even fixed the proper indentation in the source and temporarily added checkpoints where it would show me the variables during execution. If you don’t mind, I’ll just continue looking into this later today. I don’t give up easily… but I do welcome help from any experienced developer out there.

 

I also wanted to mention that I cleaned up (i.e. moved or modified) some of my code in those files involved in Client Control features so that it is more efficient and easier to understand. In so doing, I also decided to remove the DISABLE_SCREENSHOTS setting, so that it is in line with how Client Control combines the disabling of file transfers along with screenshots.

 

In regards to how Client Control manages the broadcasting option, I feel it should be all inclusive. There are 3 different ways to use the broadcast feature on the client. It isn’t logical to allow the Client Control plug-in to manage only 2 out of 3 possibilities. So it was probably an over site when it was originally written and the description needs to be clarified. It would also then match the way it is controlled on the client side with the DISABLE_BROADCAST_MENU_ITEM setting which covers all the bases. Can you please file a JIRA ticket for this? Would the ticket be for Spark or Openfire?

 

So I’ll continue to clean up the code and do more thorough testing today. When I get to a point where I feel I can submit a new PR, I’ll file it under SPARK-1822. 

 

 

 

Regards,

 

Michael

wrooot:
Yes, you ca do a new PR for 1822 (i have reopened it).
I have also filed one for Client Control (will do this myself) https://issues.igniterealtime.org/browse/OF-1211
And new one for fixing broadcasting restriction https://issues.igniterealtime.org/browse/SPARK-1827

BittyByte:
I'm getting closer to fixing this issue (SPARK-1516) and will work on it 
later today. Will let you know how it goes...

Sent with AquaMail for Android
http://www.aqua-mail.com

PR #252, opened 2016-11-02
SPARK-1827 - Shouldn't allow broadcast to selected users when it is disabled in Client Control

BittyByte:
When Broadcasting is disabled in Client Control, Spark only removes Broadcast option from in the main menu and in context menu of a group. But users are still able to select a group of users and send a broadcast to them.

### DIFF
--- a/src/java/org/jivesoftware/resource/default.properties
+++ b/src/java/org/jivesoftware/resource/default.properties
@@ -109,11 +109,8 @@ DISABLE_AVATAR_TAB = false
 # If true, disable the "Plugins" menu item
 DISABLE_PLUGINS_MENU_ITEM = false
 
-# If true, disable the option to transfer files
+# If true, disable the option to transfer files and images
 DISABLE_FILE_XFER = false
-
-# If true, disable the option to take screenshots
-DISABLE_SCREENSHOTS = false
 
 # If true, disable the option to remove contacts and groups in roster
 DISABLE_REMOVALS = false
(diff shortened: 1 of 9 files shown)


wrooot:
I will merge this PR now, but next time we should separate actual code changed and formatting improvements (wasn't expecting so many changes:)), so it would be easier to pin point regressions.

BittyByte:
Didn't know it's ok to submit a separate PR for each formatting change. 
Will know for next time. Sorry.

I'll submit a PR for SPARK-1822 later today. By the way, I'm still stumped 
about the avatar editing issue, but will read through the code today.

Thanks and have a good day.

Sent with AquaMail for Android
http://www.aqua-mail.com

wrooot:
We usually don't mind a few formatting changes in a PR. But when there are so many of them it is hard to find the actual changes, so i think it is better to separate them. We could file a ticket "Fix code formatting" or like that.

I have merged your PR. Btw, it seems that you have already submitted changes for 1822 - https://github.com/igniterealtime/Spark/commit/8beb0b9
At least there is no separate screenshot setting and disabling file transfer removes screenshot button as well. Unless you want to add additional changes to that.

BittyByte:
I did try to submit the PR for SPARK-1822 earlier tonight, but wasn't sure 
that it actually went through since I didn't see the screens where I can 
add a comment, etc. I figured it was either canceled or placed on hold 
until the previous PR was merged first. I assume PRs need to be merged in 
order.




Sent with AquaMail for Android
http://www.aqua-mail.com

wrooot:
It just was attached to your parent open PR, which i have already merged. So everything should be good now. Will close 1822 as well. Thanks.

BittyByte:
That makes sense now. Thank you for clarifying how the process works. I'll 
know for next time.

And thanks again for your patience. I'm learning more every day.

Sent with AquaMail for Android
http://www.aqua-mail.com

PR #253, opened 2016-11-04
sparkmeet plugin - initial release

deleolajide:
As suggested, here is a PR to add the plugin into the Spark main repo. 

I am not sure how this will be built in the final packaging of Spark as it depends on [Electron v 1.4.5](https://github.com/electron/electron/releases/tag/v1.4.5) runtime files to be included in the jar file during the build depending on what platform target is chosen. 

We could go "all", but that creates a large plugin (over 200MB) which causes Spark to takes over a minute to load. In any case, I have provided targets in the build.xml file for windows, linux, osx and all which can be used to create slimmer files for specific OSes


wrooot:
Lets get @guusdk and @akrherz on board with this. What do you thinks is the best approach? Will we bea ble to tune Bamboo to produce artifacts with this plugin?

wrooot:
Bundling is not a good idea currently as this plugin is relying on Ofmeet version which hasn't been released yet and which in turn is probably relying on alpha version of Openfire which hasn't been released yet and won't be for a long time.

deleolajide:
>this plugin is relying on Ofmeet version which hasn't been released 

Not exactly. It needs Ofmeet version 0.3.28 which is now available on Github.

There is a new version of Openfire Meetings (0.4.x) that @guusdk is working on. That is probably the one you are referring to.

guusdk:
Given that the Electron-based solution is likely going to be replaced by another technology (as Dele just mentioned to me via chat), I'd not get into the trouble of generating platform-specific plugin distributions. If the 1-minute delay is a one-time delay, that seems acceptable to me. People that are annoyed enough by this can always, manually, build a platform-specific plugin themselves.

wrooot:
What? New solution again? It seems to be changing rapidly :) Yes, i was wrong to say it is relying on alpha version of Openfire, but 0.3.28 is still not officially available (as an installable plugin).

When i tried to load the sparkmeet.jar (before Electron i think) in my Spark, i haven't noticed any delay in loading. But, having to bundle a 200 MB plugin for 70-40 MB application seems a bit as an overkill. Especially for those who don't need video chat (yeah, it seems almost everyone wants it though). It will take longer to build and download (say when testing new builds) and distribute (when updating clients in the network). Our servers also will take larger toll. Of course, those a not major issues in the end. If it is easier and more convenient, we can probably bundle the all-platforms plugin. I have a poll about this in the forums btw.

deleolajide:
Sorry for the confusion on this. 

SparkMeet is an improved version of the old Openfire Meetings plugin for Spark. It does exactly the same thing as before. No new features. It is **ready to be merged** and only requires additional work after merging for packaging if you decide to make it part of the core plugins.

The use of JxBrowser has now shifted focus to the new Pade project I have created at Ignite Realtime. This project will experiment with giving Spark a new HTML5 facelift using JxBrowser. It will also look at making Openfire Meetings more scalable and distributed by moving Jitsi Videobridge to the Spark client and retaining the Focus user on Openfire.

wrooot:
Dele said "The standard Spark build uses the default build.xml. I have created several versions. You simply replace build.xml what whatever you decide as the default action/behaviour. To keep it simple, lets not add it to core plugins and keep all as the default build.xml and let folks build custom platform builds depending on their needs."

Still need to clarify. If i just merge it now, how will it affect Spark? Will it just build it and add to the main package along with other plugins (flashing, roar, etc.)?

deleolajide:
> Still need to clarify. If i just merge it now, how will it affect Spark? Will it just build it and add to the main package along with other plugins (flashing, roar, etc.)?

It will have no effect on Spark and only available for anyone to build the plugin manually themselves.

It should be added to the main folder of plugins. It is not on the list of plugins in top level build.xml and will not be added to the plugins distributed with Spark, but can be build manually by running ANT on the build.xml file provided. I have also provided batch files to build for win32 and win32 & win64

wrooot:
OK, lets give it a try.

wrooot:
@akrherz the build was fine, but Mac job is in paused state. It was fine a few days ago while doing minor PRs. This one is big, so it is causing some problems probably (Openfire Mac build is also paused for a very long time). I want to apply some changes and release a new version in a few days, so i need Mac artifact working.

akrherz:
@wrooot I restarted bamboo agent, which appears to have fixed it

wrooot:
yeah, both Openfire and Spark build are fine now, thanks

PR #270, opened 2016-12-09
Spark 1825

Alexander198961:
created init configuration for save broadcast history but maybe need change some web ui logic


wrooot:
Alexander, i have merged this PR and i have added a few issues as a comment here https://issues.igniterealtime.org/browse/SPARK-1825

PR #272, opened 2016-12-12
Spark 1825

Alexander198961:
fixed:
History is saved into Spark profile's root folder and history is the same for every xmpp user logged on that machine with same OS user. Granted, in most cases this is sufficient. But some users might have two separate xmpp accounts for separate tasks and would like to have separate history of broadcasts. So ideally history should be moved to user\user@server\transcripts and should only open a history of a particular xmpp user (minor).

Broadcast history window is always positioned in the upper left corner of the screen and it doesn't save its new position when you drag the window (minor).

Adding a date before the time


Alexander198961:
will resolve conflicts

wrooot:
I see that it still conflicts. Will you be able to fix those conflicts before Friday? I want to give this feature a real life test, so i was wondering should i use the current version or wait an updated one.

Alexander198961:
need to test it after resolution conflict

Alexander198961:
Done

wrooot:
I've added my own little touch here https://github.com/igniterealtime/Spark/pull/273

PR #276, opened 2016-12-28
SPARK-1852: Add anonymous SASL login support

volatilevar:
- Add a "login anonymously" checkbox to the login dialog
- Support anonymous login using Smack
- Minor format/typo fixes

### DIFF
--- a/src/java/org/jivesoftware/LoginDialog.java
+++ b/src/java/org/jivesoftware/LoginDialog.java
@@ -158,7 +158,7 @@ public class LoginDialog {
             mainPanel.add(loginPanel,
                     new GridBagConstraints(0, 2, 2, 1,
                             1.0, 1.0, GridBagConstraints.NORTHWEST, GridBagConstraints.BOTH,
-                            new Insets(0, 0, 0, 0), 0, 0));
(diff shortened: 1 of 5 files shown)


wrooot:
I see that you have already added the default properties. Nice ;)

wrooot:
There is one problem though. The checkbox is only shown first time. Once its value is written into spark.properties file (either true or false) it is not shown on the Login screen and even deleting it from the file is not helping for some reason.

volatilevar:
Thanks for merging my changes in.

I tried but could not reproduce this problem by editing the "ccAnonymouslogin" key in spark.properties. When ccAnonymousLogin is set to true, I can always see the checkbox when Spark launches. Could you please let me know the steps you followed?

wrooot:
I understand what's happening. You've already built in support for Client Control plugin for this checkbox. But the plugin itself hasn't been updated yet, so it returns false by default (actually nothing and it is considered as boolean false). I have already contacted the person who did recent changes to that plugin. Or i will do the update myself as it seems like a simple addition.

Thanks again ;)

volatilevar:
Glad that you found the cause. Thanks!

PR #287, opened 2017-01-13
SPARK-1860: Resolve MUC nickname conflict.

guusdk:
When joining a MUC room, allow a user to change its nickname, when the initial nickname is already in use.


wrooot:
Works fine, but i wonder why the warning message only translates OK, but Cancel stays not translated.

PR #292, opened 2017-01-22
Update spark_i18n_de.properties

Atalanttore:
I've translated last strings, removed anglicisms, and added punctuation marks.


wrooot:
Thanks

wrooot:
Had to revert due to bad encoding.

PR #295, opened 2017-01-26
SPARK-1867 hand picked all actual translations

wrooot:


wrooot:
Moving all the actual translations with correct encoding from https://github.com/igniterealtime/Spark/pull/292

PR #299, opened 2017-02-06
Reconnection delay setting

speedy01:
This is an attempt to make the re-connection delay configurable, with a 10 sec default..  While doing so, removed the reconnection code from the ReconnectPanel(s) UI, since its now being handled by the reconnection manager.


wrooot:
https://issues.igniterealtime.org/browse/SPARK-1877

wrooot:
Just to confirm. This lets you set a delay between reconnection attempts? Also, it doesn't save new value, always set to 10 seconds.

speedy01:
yes..it should be saved. looks like my last commit (which was suppose to remove the invalid imports), I accidentally removed the line of code that should have saved the value!  I'll fix that!

PR #300, opened 2017-02-06
SPARK-1811: allow partial 'version' results and include progress indicator

guusdk:


wrooot:
Still can't get client version of another Spark 2.8.x client. Shows loading indicator though.

PR #303, opened 2017-02-07
SPARK-1877-Fix Save option

speedy01:


wrooot:
It saves the setting now. But it is only applied after you restart Spark. maybe we should add "(restart required)" to the name of the setting, if it can't be done other way.

speedy01:
I think we'll need to add the "restart required". Let me know if you'd like me to add it or if you would like to.

wrooot:
Nah, will add it myself. Still need to add LT translation :)

PR #317, opened 2017-03-17
SPARK-939 Stretch and move Find dialog

Alameyo:


wrooot:
Also, what do you mean by "top-right"? Would be best to still have it centered, but moved to the top of the roster and stretching it down to have more space to show search results. Unless it now covers the whole roster.

Alameyo:
When it is on top-right and you stretch Spark window wider then find contact panel would not cover your status and list of contacts, also there is much unused space on right. When you use narrow, non strechted window then there will be no difference.

wrooot:
I guess will have to see it to fully understand how it looks.

wrooot:
I have applied the PR and tested the new build. I don't like that Find dialog appears on top of the toolbar. Ideally it should still pop inside of the roster, but at the top of the roster. And i would still like it to be centered. Aesthetic is also important in software :) And when you search it is not that important for you to see the contacts. Another thing i have noticed is that it remembers the size of your first query (say it was 15 results) and if you delete your first query and try another one the size of the results window will stay the same. So if your first query was 2 contacts only, the window will stay of 2 lines size and if your second query (deleting the name and typing another one) is larger it will still show 2 lines and the scrollbar.

Maybe i'm asking too much and it is hard to make it truly dynamic and stretching to the whole size of the roster (no matter what size of your roster window is) depending on what you type and how many results are. At least it could always use the set maximum size to display results (~20 lines) and then show a scrollbar and find dialog appearing at the top of the roster.

core/src/main/java/org/jivesoftware/spark/component/JContactItemField.java
```diff
@@ -176,7 +177,11 @@ private void showPopupMenu() {
 
         if (validItems.size() != 0 && !popup.isVisible()) {
             popup.pack();
-            popup.setSize(textField.getWidth(), 200);
+            if(validItems.size()*16<=SparkManager.getMainWindow().getHeight()-SparkManager.getSearchManager().getSearchServiceUI().getHeight()-SparkManager.getMainWindow().getTopToolBar().getHeight()-SparkManager.getWorkspace().getStatusBar().getHeight()-48){
```
guusdk: For future reference: lets not make lines this long, because it makes it hard to read what's going on.

PR #320, opened 2017-03-19
SPARK-939 Stretch and move Find dialog

Alameyo:
Requested changes.


wrooot:
image
Can it be moved into marked place? So it won't appear on top of user's toolbar, but in the roster itself.

Alameyo:
Oh, now I see that avatar makes a difference. Do you want it still appearing after ctrl+f or as permanent part of the roster?

wrooot:
No, it should only appear when you press Ctrl-F. There is already a Search window at the bottom, so it would be confusing to have two search windows at once.

PR #321, opened 2017-03-20
SPARK-1875: Spark history search should show context

johnyleebrown:
Fixes SPARK-1875

### DIFF
--- a/core/src/main/java/org/jivesoftware/sparkimpl/plugin/transcripts/HistoryTranscript.java
+++ b/core/src/main/java/org/jivesoftware/sparkimpl/plugin/transcripts/HistoryTranscript.java
@@ -35,6 +35,7 @@ import org.jivesoftware.spark.util.log.Log;
 import org.jivesoftware.sparkimpl.settings.local.LocalPreferences;
 import org.jivesoftware.sparkimpl.settings.local.SettingsManager;
 import org.jxmpp.util.XmppStringUtils;
+import static org.apache.commons.lang3.StringEscapeUtils.unescapeHtml4;
 
 /**
  * This class represents the history transcript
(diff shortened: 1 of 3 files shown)


johnyleebrown:
Found out that highlighted search text stays highlighted after the search is over. Fix is coming.

core/src/main/java/org/jivesoftware/sparkimpl/plugin/transcripts/HistoryTranscript.java
```diff
@@ -262,7 +260,8 @@ private synchronized void display() {
 			token.acquire();
 
 			if ((searchFilteredList.size() > 0) && (pageIndex <= searchFilteredList.size())) {
-				builder.append(buildString(searchFilteredList.get(pageIndex-1).getMessages()));
+                builder.append(org.apache.commons.lang3.StringEscapeUtils.unescapeHtml4(
```
guusdk: Could you use an import statement? That way, you can do ``StringEscapeUtils.unescapeHtml4`` instead of ``org.apache.commons.lang3.StringEscapeUtils.unescapeHtml4`` - which makes the code more readable.

johnyleebrown: Sure.

guusdk:
Your fix appears to highlight the text that is being searched for in the results, which is a good improvement, but not quite what SPARK-1875 was intended for.

When searching in your history, it is often useful to have a few lines of the conversation before and after the line that matches the search. That way, you can easily see what you were talking about. Adding that context was the original intention of SPARK-1875. 

I'd definitely keep the highlight if you can make it work, because that's also a good improvement, but can you include a bit of the chat context too?

johnyleebrown:
@guusdk i understand now) i'm fixing the highlight and then i'll get to the context part.

wrooot:
To be fair, Guus, i only now understood what you meant in that ticket :)

johnyleebrown:
@guusdk fixed highlighting, if the issue regarding highlighting searched text existed, i could created pr for that.

PR #323, opened 2017-03-21
SPARK-1635: History viewer adds spaces between messages when changing a view

johnyleebrown:
Fixes SPARK-1635

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -140,6 +140,11 @@
             <version>3.1</version>
         </dependency>
         <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.1</version>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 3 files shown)


guusdk:
Excellent to see more fixes from you! Thanks!

I'm noticing that this PR also includes the fixes for SPARK-1875, even though you've created a new branch. I think you created the branch from a source that already included the SPARK-1875 fix, instead of just before. This is not ideal (given that the work for SPARK-1875 has not yet been merged in the official Spark repository).

johnyleebrown:
@guusdk thanks. ok, then i'll finish with context highlighting.

guusdk:
Finishing is always good, but I'd like you to try and see if you can make the two PRs be independent of each-other too. That will allow you to work on both efforts in parallel - which is something that you probably want to do in the future a lot.

johnyleebrown:
@guusdk ok, sure.

PR #325, opened 2017-03-21
SPARK-1635

johnyleebrown:
SPARK-1635


wrooot:
Guus, do you have any comments here?

PR #326, opened 2017-03-21
SPARK-1875: Spark history search should show context

johnyleebrown:
Fixes SPARK-1875:
Added one line before and after searched message.
Added yellow highlighter over searched message.

### DIFF
--- a/core/src/main/java/org/jivesoftware/sparkimpl/plugin/transcripts/ChatTranscript.java
+++ b/core/src/main/java/org/jivesoftware/sparkimpl/plugin/transcripts/ChatTranscript.java
@@ -71,20 +71,29 @@ public class ChatTranscript {
      * @return the messages that included search keywords.
      */
     public List<HistoryMessage> getMessage(String text) {
-    	if(text == null || "".equals(text)) {
-    		return messages;
-    	} else {
-	    	List<HistoryMessage> searchResult = new ArrayList<>();
-	    	for(HistoryMessage message : messages) {
-	    		// ignore keywords' case
-	    		if( message.getBody().toLowerCase().contains( text.toLowerCase() ) ) {
-	    			searchResult.add(message);
-	    		}
-	    	}
-	    	return searchResult;
-    	}
+        if (text == null || "".equals(text)) {
+            return messages;
+        } else {
+            List<HistoryMessage> searchResult = new ArrayList<>();
+            int iter = 0;
+            HistoryMessage previous = null;
+            for(HistoryMessage message : messages) {
+                iter++;
+                if (message.getBody().toLowerCase().contains(text.toLowerCase())) {
+                    if (previous != null) {
+                        searchResult.add(previous);
+                    }
+                    searchResult.add(message);
+                    if (iter + 1 != messages.size()) {
+                        searchResult.add(messages.get(iter + 1));
+                    }
+                }
+                previous = message;
+            }
+            return searchResult;
+        }
     }
-    
+
     /**
      * Clears the Message History if its not needed anymore
      */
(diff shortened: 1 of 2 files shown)


wrooot:
Ok, this time i will wait for @guusdk to review this PR :)

core/src/main/java/org/jivesoftware/sparkimpl/plugin/transcripts/ChatTranscript.java
```diff
@@ -71,20 +71,33 @@ public void addHistoryMessage(HistoryMessage entry) {
...
+            List<HistoryMessage> searchResult = new ArrayList<>();
+            int iter = 0;
+            HistoryMessage previous = null;
+            for(HistoryMessage message : messages) {
+                iter++;
+                if (message.getBody().toLowerCase().contains(text.toLowerCase())) {
+                    message.setBody(message.getBody().toLowerCase().replaceAll(text.toLowerCase(),
+                            "<span style=\"background-color: #ffff00\">" + text.toLowerCase() + "</span>"));
+
+                    if (previous != null) searchResult.add(previous);
```
guusdk: Please always wrap if-statements, even if they are just one line. Not wrapping them makes for easy mistakes.

guusdk:
I love how the new search looks! It's quite an improvement! However, I did find a problem.

Your fix is based on HTML. To make your fix visible, you need to render the log as HTML (you use ``unescapeHtml4``) This renders _all_ HTML that's in the chat, which is not good.

Imagine that people have been talking about HTML, copy/pasting snippets. See this simple example, where I used a simple ``span``. As expected, the span shows (as text) in my history of Spark that does not have your fix:
spark-before

However, after your fix, the ``span`` gets rendered! That's not good (There are also a lot of duplicate lines now - I don't know if that relates to your fix, or if that's caused by a separate issue):
spark-after

Things get *really* bad after I searched for a very common needle (I used the letter 'a'). After doing that, I did my original search again, which now showed as this:
spark-after-eek
I can only imagine that there's some dangling HTML in my previous search results. I'm not sure. In any case, this should not happen. :)

johnyleebrown:
@guusdk @wrooot 
I removed the part about the highlighting for now.
I have a suggestion to implement first the "context" - one line before and after searched message. 
The highlighting method have to be completely redone and i think i could do it after the gsoc proposal submission due to lack of free time.

guusdk:
That's totally understandable. I think the commons-lang3 dependency can now also be removed, right?

johnyleebrown:
@guusdk yes, done.

core/src/main/java/org/jivesoftware/sparkimpl/plugin/transcripts/ChatTranscript.java
```diff
@@ -71,20 +71,27 @@ public void addHistoryMessage(HistoryMessage entry) {
...
+        if (text == null || "".equals(text)) {
+            return messages;
+        } else {
+            List<HistoryMessage> searchResult = new ArrayList<>();
+            int iter = 0;
+            HistoryMessage previous = null;
+            for(HistoryMessage message : messages) {
+                iter++;
+                if (message.getBody().toLowerCase().contains(text.toLowerCase())) {
+                    if (previous != null) searchResult.add(previous);
```
guusdk: Please always wrap if-statements, even if they are just one line. Not wrapping them makes for easy mistakes.

johnyleebrown: Okay, sure.

johnyleebrown:
@guusdk done.

wrooot:
I have reopened the ticket as it doesn't search now in some cases.

PR #329, opened 2017-03-28
SPARK-939 additional dialog position adjustment

wrooot:


wrooot:
Made additional change to https://github.com/igniterealtime/Spark/pull/320 and moved Find dialog a bit lower. Though it seems that different skins have different positions and sizes of the main toolbar, so with JTattoo it will be a bit too low, but it is still better than it was before.
image

Alameyo:
I thought you want yet something a little bit different and I was cracking head about it. I will add one more PR so it will be in this position regardless of most of the skins and regardless of avatar, because now if you havent got avatar position is again wrong (I forgot about avatar earlier, but without avatar it was good).

PR #330, opened 2017-03-28
Spark 939 dynamic position

Alameyo:
Now the find dialog should adjust position to the size of the status panel.


wrooot:
It is now in the right position with JTattoo skins, but it appears on top of the avatar and toolbar for Substance and LittleLuck skins. Not sure if it is possible to make it truly dynamic for all types of skins Spark has. Skins are calculating toolbars position and size differently somehow. JTattoo is not coloring the menu itself (File, Actions, etc.). Substance does, so maybe for Substance toolbar is higher than for JTattoo.

Alameyo:
I don't know which solution will be better for now, but for skins displayed properly with this change, it is displayed properly regardless of avatar. I will try make it dependent on skin, but I am not sure about it.

wrooot:
I have played a bit with it and it seems there is no ideal solution (also LittleLuck skin has ~3 pixels off the position). So i have decided to add more lines, so it would be as high as possible for any skin (with or without avatar) and won't be on top of the toolbar. Yeah, it will be a bit lower for JTattoo skins. But it is still better than it was in the middle of the roster. And in future we might change the default skin anyway (JTattoo looks dated and ugly :)). Thanks for you help.

https://github.com/igniterealtime/Spark/pull/331

PR #332, opened 2017-03-31
Reload vCards in pre-2.8 schema

LookAtThatThing:
After updating from 2.7.6 to 2.8.3 recently our vCards stopped showing information and would no longer update.

Looking at vCard xml it looks like the schema changed after 2.7 which added an extra root node causing files saved in 2.7 not to parse correctly. As the file exists Spark won't fetch the vCard, but because the parser can't get the timestamp (or any other data), it won't refresh it either causing it to get stuck and appear empty. This change regrabs the vCard in that scenario, saving it in the new schema in the process.

### DIFF
--- a/core/src/main/java/org/jivesoftware/sparkimpl/profile/VCardManager.java
+++ b/core/src/main/java/org/jivesoftware/sparkimpl/profile/VCardManager.java
@@ -785,6 +785,12 @@ public class VCardManager {
                     addToQueue(jid);
                 }
             }
+            else
+            {
+            	//vCard xml schema changed 2.7->2.8
+            	//if we have a valid file, but no valid timestamp - reload the vCard (probably in old style) 
+            	addToQueue(jid);
+            }
 
             addVCard(jid, vcard);
             in.close();


guusdk:
Thanks for your contribution!

I have a couple of concerns:

- Was the VCard syntax supposed to change in the first place?
- The suggested change won't work immediately - it will cause the 'incorrect' VCard to be returned, although it will mark it for refreshment. I suppose that, depending on timing, it will cause the correct data to appear whenever the VCard is requested again.
- If we have invalid VCard data on file, the parser should fail and throw an exception. The file probably should be deleted at that point too.

speedy01:
wroot and I had a brief discussion on this a vcard caching a while back, and wonder if its even worth cache at all?  Or maybe only cache to memory?  Thoughts?

guusdk:
Some VCards can (should not be, but can) be huge. Even we only recently started to shrink avatars that were to big. An offline cache of sorts remains a sensible thing, I'd say.

guusdk:
I've created an alternative PR, which could replace this one.

See https://github.com/igniterealtime/Spark/pull/339

PR #333, opened 2017-03-31
Quick Start Guide for IntelliJ

speedy01:


guusdk:
Thanks for this! The only thing that I'd change is the first batch of images: I would create two rows of three images, and put some text in between.

wrooot:
I will leave it to Guus to review and commit ;)

PR #334, opened 2017-04-04
Don't declare transitive Smack dependencies

Flowdalic:
The artifacts smack-resolver-javax and smack-sasl-javax are transitive
dependencies of smack-java7. That is all we need to depend on.


guusdk:
That's sensible, thanks.

wrooot:
This doesn't require JIRA issue?

guusdk:
Nah.

PR #335, opened 2017-04-09
 SPARK-1903: Highlight searched keyword in the history search

johnyleebrown:
SPARK-1903

### DIFF
--- a/core/src/main/java/org/jivesoftware/sparkimpl/plugin/transcripts/HistoryTranscript.java
+++ b/core/src/main/java/org/jivesoftware/sparkimpl/plugin/transcripts/HistoryTranscript.java
@@ -11,10 +11,16 @@ import java.awt.event.KeyEvent;
 import java.awt.event.KeyListener;
 import java.awt.event.WindowAdapter;
 import java.awt.event.WindowEvent;
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.StringReader;
+import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.*;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import javax.swing.*;
 import javax.swing.text.*;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/jivesoftware/sparkimpl/plugin/transcripts/HistoryTranscript.java
```diff
@@ -258,18 +266,47 @@ public final String buildString(List<HistoryMessage> messages){
...
+					BufferedReader buf = new BufferedReader(new StringReader(str));
+					int globalPos = 1;
+					while ((line = buf.readLine()) != null){
+						Matcher matcherTime = Pattern
+								.compile("\\(\\d\\d:\\d\\d:\\d\\d\\)\\s[a-zA-Z0-9]+:\\s")
+								.matcher(line);
+						if (matcherTime.find()) {
+							for (int i = matcherTime.end(); i + text.length() < line.length() + 1; i++) {
+								String match = line.substring(i, i + text.length());
+								if (text.toLowerCase().equals(match.toLowerCase())) {
```
guusdk: This does the same, but is somewhat easier on the eyes: ``text.equalsIgnoreCase( match )``

core/src/main/java/org/jivesoftware/sparkimpl/plugin/transcripts/HistoryTranscript.java
```diff
@@ -258,18 +266,47 @@ public final String buildString(List<HistoryMessage> messages){
...
+									window.getHighlighter().addHighlight(globalPos + i,
+											globalPos + i + text.length(), highlighter);
+								}
+							}
+						}
+						globalPos = globalPos + line.length() + 1;
+					}
+				}
+			} catch (BadLocationException | IOException ex) {
+				ex.printStackTrace();
```
guusdk: Please don't do this - always print exceptions to a Log object!

johnyleebrown:
@guusdk Done.

guusdk:
Okay, thanks. I see that you now do a Log as well as a printStackTrace, which is also done in other places of this class. In the future, we should not do that - we should never use printStackTrace, only use the Logging framework. PrintStackTrace will always output to stdout, while the logging framework is configurable (and will put errors to the logfiles). In short never use printStackTrace, always use log. :)

johnyleebrown:
@guusdk Got it, thanks)

wrooot:
It doesn't highlight text in URLs. It only highlights one instance of searched word (which is in the center of the window). I guess showing the context is messing with it (the searched word can be in the context if it was repeated a few times).
image
image

johnyleebrown:
@wrooot 

screen shot 2017-04-23 at 11 11 57

Links
screen shot 2017-04-23 at 11 14 34

wrooot:
I guess it depends on the amount of messages or some other stuff. Sometimes it works, sometimes don't. We can mark it fixed for now, unless someone comes up with an exact way to reproduce it.

PR #336, opened 2017-04-11
Transcript Window should be recomposable.

guusdk:
This commit started off as a fix for SPARK-1757, but ended up being somewhat of a rewrite of the
code that displays chat messages in the UI (the 'chat transcript window').

The original problem defines that chat messages are sometimes displayed out of order. The direct cause
of this is unknown (I suspect threading/timing issues). To work around the issue, this commit now
adds the capability to recompose the entire transcript window from memory.

A drawback of this is that a(nother) copy of the chat is now kept in memory (the 'entries' list that
is now added to TranscriptWindow). This might make Spark use more memory.

I've opted for a rewrite, as in the course of applying fixes to the existing code, I repreatedly got
lost in the jungle of reuse of styling, particularly. This commit adds new classes for all types of
entries in the Transcript Window (typically, a chat message), and makes each class responsible for
creating a UI representation of that type of entry. This removed quite a bit of boiler plate code.

As each entry now contains the original message, it's make-up as well as timestamp, the transcript
window is easily redrawn when messages start to arrive out of order. That bit is now implemented as
a responsibility of the TranscriptWindow class.

Various smaller changes were made:
- ctrl-space nickname completion in a MUC now no longer considers users on your roster (but not in the MUC)
- ctrl-space nickname completion now works for users that have no VCard (uses node-part of the JID instead)
- message color is now configurable (but still defaults to black).

### DIFF
--- a/core/src/main/java/org/jivesoftware/spark/Workspace.java
+++ b/core/src/main/java/org/jivesoftware/spark/Workspace.java
@@ -401,7 +401,7 @@ public class Workspace extends JPanel implements StanzaListener {
         }
 
         // Insert offline message
-        room.getTranscriptWindow().insertMessage(nickname, message, ChatManager.FROM_COLOR, Color.white);
+        room.getTranscriptWindow().insertMessage(nickname, message, ChatManager.FROM_COLOR);
         room.addToTranscript(message, true);
 
         // Send display and notified message back.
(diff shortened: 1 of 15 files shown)


guusdk:
Note that being able to recompose the transcript window will be handy when we want to implement https://xmpp.org/extensions/xep-0308.html

Alameyo:
I checked it and one minor thing to improve: that green wall of join/left messages was on the top of the window and now it is breaking messages in half as recived befor and after joining to room. I don't think it looks good. 
greenwallbetweenmessages

guusdk:
I'm totally not bothered by that, but I can see how others are. :) I'm not to sure there's an easy fix for this, as entries are simply displayed as the arrive. Do you see a way to fix this?

Alameyo:
Not yet, I have done SPARK-1519 with timestamps but now I have to check this changes in TranscriptWindow to remake it, so during that maybe I will see the fix :)

PR #338, opened 2017-04-12
SPARK-1909 Green blob of user join/left messages now is on top of the transcript window.

Alameyo:
New joined/left and other presence messages still will appear durning conversation, but not informations about who joined/left before we joined the room.

### DIFF
--- a/core/src/main/java/org/jivesoftware/spark/ui/TranscriptWindow.java
+++ b/core/src/main/java/org/jivesoftware/spark/ui/TranscriptWindow.java
@@ -109,7 +109,7 @@ public class TranscriptWindow extends ChatArea implements ContextMenuListener
         //boolean reorderEverything = false;
         if ( !entries.isEmpty() )
         {
-            if ( entry.getTimestamp().isBefore( entries.getLast().getTimestamp() ) )
+            if ( entry.getTimestamp().isBefore( entries.getLast().getTimestamp() ) && !(entries.getLast() instanceof CustomTextEntry) )
             {
                 Log.warning( "A chat entry appears to have been delivered out of order. The transcript window must be reordered!" );
                 //reorderEverything = true;
@@ -122,7 +122,7 @@ public class TranscriptWindow extends ChatArea implements ContextMenuListener
                                                 // Clear and refill the UI component.
                                                 try
                                                 {
-                                                    entries.sort( Comparator.comparing( TranscriptWindowEntry::getTimestamp ) );
+                                                    entries.sort( Comparator.comparing(TranscriptWindowEntry::isDelayed).thenComparing( TranscriptWindowEntry::getTimestamp ) );
                                                     clear();
                                                     for ( TranscriptWindowEntry e : entries )
                                                     {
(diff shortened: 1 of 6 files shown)


wrooot:
Suggest assigning to SPARK-1909 when merging.

Alameyo:
So:
!e.getTimestamp().isBefore(drawTime) - exclude things that were send to chat before that time, i.e. messagesEntries as well as customEntries.
!(e instanceof CustomTextEntry) - exclude CustomTextEntres (joined/left but also changed subjects and maybe some others).
|| -  MessageEntries and other will always come through because they aren't CustomEntries. CustomEntries will come only if their timestamp is after joining the room (after drawTime ).

guusdk:
Uff - could you at the very least document this in the code? I'll not remember this in two days...

Alameyo:
Ok.

Alameyo:
This piece of code is called everytime since, entries always are out of order, because Openfire sends first "archival" presence entries which we timestamp when they come and then archival MessageEntries which have delay extension. The only exception is when there are only presence entries entries in room. However I understand it might not be the case with every server or maybe even you can set this in Openfire.

wrooot:
Is this still requiring changes?

guusdk:
Yes, I feel that we need to change this. I really dislike this functionality to be added in a piece of code that ideally is never invoked. The code where it is put in was only added to correct some out-of-order scenario, that really should not occur in the first place.

Alameyo:
Yes, I need to work on this a bit more.

Alameyo:
Generally:
- now exist isDelayed() method for entries.
 - there is change in sorting in case of messages out of order. 
 - CustomEntries don't invoke sorting (since CostumMessages have no delay so that would messing things )
 - rest are mostly additional Constructors.

core/src/main/java/org/jivesoftware/spark/ui/TranscriptWindow.java
```diff
@@ -122,11 +122,11 @@ protected synchronized void add( TranscriptWindowEntry entry )
                                                 // Clear and refill the UI component.
                                                 try
                                                 {
-                                                    entries.sort( Comparator.comparing( TranscriptWindowEntry::getTimestamp ) );
+                                                    entries.sort( Comparator.comparing(TranscriptWindowEntry::isDelayed).thenComparing( TranscriptWindowEntry::getTimestamp ) );
                                                     clear();
                                                     for ( TranscriptWindowEntry e : entries )
                                                     {
-                                                        e.addTo( transcriptWindow );
+                                                    	e.addTo( transcriptWindow );
```
guusdk: This is an unintentional whitespace change, I think.

core/src/main/java/org/jivesoftware/spark/ui/TranscriptWindowEntry.java
```diff
@@ -89,4 +103,13 @@ public int hashCode()
     {
         return timestamp.toInstant().hashCode();
     }
+    /**
```
guusdk: There's some whitespace oddity here too.

core/src/main/java/org/jivesoftware/spark/ui/TranscriptWindowEntry.java
```diff
@@ -38,13 +39,26 @@
...
+    	this(timestamp, false);    	
+    }
+    /**
+     * Constructs a new entry.
+     *
+     * When displayed in context of a conversation, chat entries are ordered by time. This is why the time component
+     * of an entry is non-optional. Note that multiple entries can exist that have the same time component.
+     *
+     * @param timestamp The timestamp of the entry (cannot be null).
+     * @param isDelayed Set true if timestamp come from delay extension.
```
guusdk: I'd change the description of the `isDelayed` argument, as it should be true for any message that is 'historic'. A 'delay' extension is just one way of expressing that. There might be other ways of expressing this too.

guusdk:
Thank you!

PR #342, opened 2017-05-05
SPARK-1910 Replace (some) emojis with emoticons.

Alameyo:
Now Spark will recognize ☺ and ☹ as emoticons.


Alameyo:
It seems that Spark is unziping emoticons only during installation. To fast check if it works go to core/src/main/xtra/emoticons and unzip files to C:\Users\<Username>\AppData\Roaming\Spark\xtra\emoticons.

wrooot:
So, if you add ;) and :D we can merge it and then generate the installer and test it. Unless @guusdk wanted something else

Alameyo:
You can merge, an check if it will work with installer.

PR #343, opened 2017-05-05
SPARK-1915 -- Implement separate History settings in Client Control

BittyByte:
Currently Client Control plugin's setting "History Settings" not just hides the history settings, but also disables the history in Spark. Which might not be desired. One might need to have history, but hide settings, so users won't be able to disable it from a GUI. This change will implement two separate history features in Client Control - "History Settings" and "History Transcripts". Now an admin can select to enable/disable the history controls as well as enable/disable the logging of the history into transcript files. **NOTE: Updated version 2.1.2 of the Client Control plug-in and minimum of Spark version 2.9.0 is required for proper functionality.**

### DIFF
--- a/core/src/main/java/org/jivesoftware/resource/Default.java
+++ b/core/src/main/java/org/jivesoftware/resource/Default.java
@@ -93,6 +93,7 @@ public class Default {
     public static final String HISTORY_DISABLED = "HISTORY_DISABLED";
     public static final String HIDE_HISTORY_SETTINGS = "HIDE_HISTORY_SETTINGS";
     public static final String HIDE_SAVE_PASSWORD_AND_AUTOLOGIN = "HIDE_SAVE_PASSWORD_AND_AUTOLOGIN";
+    public static final String HIDE_LOGIN_AS_INVISIBLE = "HIDE_LOGIN_AS_INVISIBLE"; 
     public static final String MAINT_FILESPEC = "MAINT_FILESPEC";
 
     static ClassLoader cl = SparkRes.class.getClassLoader();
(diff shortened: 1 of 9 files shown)


wrooot:
It seems that i can't add a reviewer properly if he isn't in the Spark Team. But that's not required. As i have mentioned to BittyByte i will ask our GSoC student @Alameyo to review this PR. I have to explain that this is a two-part PR. There will be accompanying PR for Openfire. There is Client Control plugin for Openfire, which lets admins to control various settings of Spark. So when doing changes, they had to be done in both places (Openfire/Spark).
@BittyByte add a comment about this PR once you send a PR for Openfire, so both PRs will be linked.

BittyByte:
I’m just about to do the PR for Openfire. I’ll let you know what PR # was assigned.

BittyByte:
So just put me on the team! ;-)

 

PR #796  à “OF-1325 -- Implement separate History settings in Client Control”

BittyByte:
**SPARK-1868:** Invisible Login and Invisible presence have been added to Spark since 2.7.0 version. Some administrators might want to disable this via default.properties or via Client Control plugin. Invisible Login can already be disabled, but Invisible presence not yet. This PR resolves this issue.

wrooot:
I was referring to alameyo when talking about team :) U think team members have full access to the repo, so only a few folks are in there.

Anyway, i think i'm going to merge this, but you somehow added different commits to the same PR (probably because first commit wasn't yet commited and your changes were based on the first one). Well, all the changes are related to the same functionality, so i think it will be ok.

PR #344, opened 2017-05-30
SPARK-1926, SPARK-1927, SPARK-1928, SPARK-1929, SPARK-1938 and SPARK-1937

Alameyo:
SPARK-1926, SPARK-1927, SPARK-1928, SPARK-1929, SPARK-1938 and (here might be some changes in future) SPARK-1937.


Alameyo:
Now also SPARK 1940.

guusdk:
Let's merge this code, after you build in something that, while under construction, does *not* show the new features, without toggling a property. I think you already did most of that work with your addition to `org.jivesoftware.resource.Default` ...

Alameyo:
> You appear to be applying an MVC pattern, correct?

Yes, I added note in Javadocs about it.

> It is probably a good idea to do input validation on the argument values in constructors and methods.

Done

> ...use logging! And not to stdout please, but use the logfiles. ;)

Already I am doing that when catching exceptions. I am not sure if I should do it also somewhere else.
`catch(Exception e) { 
Log.warning("Something wrong", e)
}`

> I think your editor wraps comments/javadoc at 80 characters width? I'd change that to 120 or 160. No-one is using text terminals any more. ;)

Oh, yeah my comments/javadocs were only 80 while rest of the code 120. I changed it :)

> It might be worth to combine somehow the existing 'PKI' tab, and your new 'certificates' tab.

I thought about it, but I am not sure how it will be combined well with panel for mutual authentication that I will need to add somewhere later in project. In any case it shouldn't be a problem to combine things from PKI tab with new panels in later.

> Let's merge this code, after you build in something that, while under construction, does not show the new features, without toggling a property.

It seems that the simplest is commenting out addition of that tab, which now I have done.

core/src/main/java/org/jivesoftware/spark/ui/login/LoginSettingDialog.java
```diff
@@ -86,7 +86,7 @@ public boolean invoke( JFrame owner )
         }
         if ( !Default.getBoolean(Default.CERTIFICATES_MANAGER_DISABLED))
         {
-        	tabbedPane.addTab( Res.getString( "tab.certificates" ), certManager );
+        //	tabbedPane.addTab( Res.getString( "tab.certificates" ), certManager );
```
guusdk: Although this will obviously do the trick, isn't the purpose of this `if` statement to enable/disable things, from the value of `CERTIFICATES_MANAGER_DISABLED`? Shouldn't we use that?

Alameyo:
Now i think it's fine.

guusdk:
Excellent. Let's not forget to switch this back after you're done with the project :)

wrooot:
This looks a bit weird (spacing between options)
image
I understand that Advanced window has been stretched to accomodate larger Certificates tab. Not sure if something can be done about it. This looks different from what is used through out the application (spacing).

Alameyo:
I didn't done nothing more than adding certificates tab, what would be visible for users with old panels. It seems I can reduce a bit that spacing between options. As I am checking older versions it doesn't appear for me as this dialog become stretched, but I can try to size it down.

wrooot:
Having an empty spot at the bottom is not pretty too, but at least it would be consistent with other parts of  Spark's GUI (say in Preferences, many windows have a few options and all of them are stacked at the top of the window).
This is the not stretched version for comparison
image

Alameyo:
That's how it looks for me, so it is even more stretched here but it was this way all the time before I started adding Certificates tab. I am not sure yet why you and me see it different but I will try to investigate that, maybe I will also allow users to resize this.

obraz

guusdk:
The UI will look very differently, depending on what OS you're using, and what theme/style gets enabled. 

Let's go with @wrooot 's suggestion, and "anchor" the content (but not the buttons on the bottom) to the top-left corner (instead of 'stretching' it over the entire height). That should give you 'empty space' between the checkboxes and the Ok/Cancel/UseDefault buttons, if the window is larger than the content.

PR #345, opened 2017-06-09
SPARK-1932, SPARK-1930, SPARK-1935, SPARK-1933

Alameyo:
And also some minor visual changes.


guusdk:
Oh, and these commits didn't include a single line of javadoc, I think. You don't need to overdocument stuff, but some text that describes what the intended purpose of code is, is good.

Alameyo:
That's weird, it should open on double click on the certificate in the table. Is your truststore setup and table show certificates properly? I can add edit button that would do the same as double clicking and also add certificate button. I will also add javadocs.

guusdk:
I don't have done anything - so I might be missing a truststore (the table is empty for me). Does Spark ship with a default truststore? If not: perhaps it should? Or, if we don't - perhaps we should use the truststore that is part of the JRE, and only store the exceptions in our truststore? If unsure, let's discuss with @wrooot. In any case, it is desirable for this to work with a standard installation of Spark (the user should not first have to manually create stores - most of the users won't understand that).

guusdk:
Also, I think you should rebase the branch, as there are some conflicts with the current 'master'.

wrooot:
Well, i don't know how it should work technically and not sure what is possible, but ideally it should just work out of the box. If it needs a new truststore, then maybe Spark should automatically create it when first launched.

Alameyo:
You have to set up truststore in PKI tab. There are 3 different types of keystores which are Java keystore, AppleKeychain and PKCS#11.

 By default Java comes with default keystore in java installation folder Java\jdk1.8.0_112\jre\lib\security. You can point to that in PKI tab, provide password which by default is "changeit" and my new panels will show content of this store. You can also try to add web browser's keystore or Openfire's but I didn't tried it yet. 

This PKI way of setting up external keystores was there before I started my project but I think it is possible to ship Spark with keystore. My concern then is which certificates we want to add to default truststore?

I still have to add some dialog in case user didn't set up truststore and I have JIRA issue for it.

Alameyo:
Added buttons and Javadocs. I will rebase after merging this, as this conflicts are annoying.

guusdk:
Thanks! On UI/UX, we can debate endlessly I think, which I'll try to prevent doing. :)

There are a copule of things that we should fix (and a lot that I think would be good suggestions, but I'll not wait for those before merging this PR).

certificate-dialog

On this screen:

- Add an 'OK' button to close the window
- Make the scrollpane be as high as the entire window.
- When you open the dialog, the page should be 'scrolled up'

Future suggestions:
- Use tabs! For instance, a tab for the issuer, another tab for the subject, etc, etc. You can then nicely layout each field, and we can, in the future, add more tabs without breaking the UI of the entire thing!
- The signature value now seems to be the toString result of a byte array? That's probably wrong.
- On the screen that lists all certificates: make it possible to sort by column!

guusdk:
Excellent, thanks! Can you rebase so that I can merge this without conflicts?

Alameyo:
Just I am not sure what do you mean by tabs. Do you mean to make it as JTabbedPanel with tab for every certificate attribute? Also scrolling up still doesn't work well but I think you can merge it, I will look into scrolling later.

guusdk:
Well, perhaps not *every* certificate attribute - related attributes could be placed on a same tab. But yeah.

Using tabs would give you more space to nicely layout the content of a certificate. For instance, you now have one text field for the subject attribute. The subject is a comma-separated list of values: it would be nicer to have a nice label + value for each of the values. That's almost not do-able in the one screen that you have now (unless we accept a lot of scrolling), but on a new tab, you create extra visible space that you can easily play with.

Also, when in the future you want to change something (add, remove or update how a particular attribute is shown to the end-user), then you only have to change one tab - changes won't affect all the other stuff.

Tabs are used in many places already, so it should also give you a nice, consistent look.

Alameyo:
Ok. I was intending to separate that fields like common name, organization unit, country name but didn't done it yet. Also so far I wasn't thinking about tabs for that but it might be good idea.

PR #346, opened 2017-06-18
SPARK-1944 

Alameyo:
Implementation of uploading certificates into Truststore. I also changed some of access modifiers to private.
I am not sure if I am doing good use of text field next to choose file. Maybe it would be better if I put there text "Certificate has been uploaded" or delete that text field at all.


core/src/main/java/org/jivesoftware/spark/ui/login/CertificatesManagerSettingsPanel.java
```diff
@@ -152,7 +157,33 @@ private void showCertificate() {
...
+			try {
+				fileField.setText(file.getAbsolutePath());
+				String certAlias = (String) JOptionPane.showInputDialog(null, Res.getString("dialog.provide.alias"));
+				if (certAlias == null) {
+					throw new IllegalArgumentException();
+				}
+				certControll.addCertificateToKeystore(file, certAlias);
+			} catch (KeyStoreException | CertificateException | NoSuchAlgorithmException | IOException ex) {
+				Log.error("Cannot upload certificate file", ex);
+			} catch (IllegalArgumentException ex) {
```
guusdk: An IllegalArgumentException is a runtime exception - which is an exception that is intended to be used when something occurs that is considered to be a programming error (not user in put error). You should not use these for things like User-input validation. For that, use a checked exception instead.

Runtime exceptions are generally not 'catched', but bubble their way up all the way to a very core class, which generically processes them in a big 'catch-all' block.

For some background, read https://docs.oracle.com/javase/tutorial/essential/exceptions/runtime.html

core/src/main/java/org/jivesoftware/spark/ui/login/CertificatesManagerSettingsPanel.java
```diff
@@ -152,7 +157,33 @@ private void showCertificate() {
...
+				Log.warning("Certificate alias cannot be null");
```
guusdk: Please, always log the exception if you have one. Instead of `Log.warning("blah")` do `Log.warning("blah",ex)`

core/src/main/java/org/jivesoftware/sparkimpl/certificates/CertificateController.java
```diff
@@ -91,6 +94,20 @@ public CertificateController(LocalPreferences localPreferences) {
 		}
 	}
   
+	public void addCertificateToKeystore(File file, String alias) throws KeyStoreException, CertificateException, NoSuchAlgorithmException, IOException{
+		if(file == null){
+			throw new IllegalArgumentException();
+		}
+		CertificateFactory cf = CertificateFactory.getInstance("X509");
+		InputStream inputStream = new FileInputStream(file);
```
guusdk: You can use `try-with-resources` here, I think.

Alameyo:
I have changed it as you requested. I am not sure what I should do about that IlegalArgumentException. Maybe I should make one, combined dialog asking about alias and showing certificate but I also want it easy distinguishable and not hidden somewhere among certificate options (which there might me more soon).

core/src/main/java/org/jivesoftware/spark/ui/login/CertificateDialog.java
```diff
@@ -198,9 +206,23 @@ public CertificateDialog(LocalPreferences localPreferences, CertificateModel cer
...
 
+	public CertificateDialog(LocalPreferences localPreferences, CertificateModel certificateModel) {
+		this(localPreferences, certificateModel, null);
+	}
+
 	@Override
 	public void actionPerformed(ActionEvent e) {
 		if(e.getSource() == okButton){
+			//controller should be passed to this class only if there is need to modification content of Keystore.
+			System.out.println(certControll);
```
guusdk: System.out? Booh!

core/src/main/java/org/jivesoftware/spark/ui/login/CertificatesManagerSettingsPanel.java
```diff
@@ -46,7 +47,7 @@
 	private final static Insets DEFAULT_INSETS = new Insets(5, 5, 5, 5);
 	private final LocalPreferences localPreferences;
 	private CertificateController certControll;
-	private JTable certTable = new JTable();
+	private static JTable certTable = new JTable();
```
guusdk: Why did you make this static? It results that all instances of this Panel will share the same instance of the table. I'm not sure if that's desirable.

core/src/main/java/org/jivesoftware/sparkimpl/certificates/CertificateController.java
```diff
@@ -58,6 +63,8 @@ public CertificateController(LocalPreferences localPreferences) {
 				case 2:
 					return String.class;
 				case 3:
+					return String.class;
+				case 4:
 					return Boolean.class;
 				default:
 					return String.class;
```
guusdk: You have exactly 5 columns now, correct? The `default` block should never be executed, I think?

If that's correct, then instead of running a value, you should throw an unchecked exception, as it would clearly be a bug in the code if that state was reached!

core/src/main/java/org/jivesoftware/spark/ui/login/CertificatesManagerSettingsPanel.java
guusdk: Perhaps we can prevent having an exception completely. When a user dies not fill out a correct alias, we could simply ask the user again, with some kind of warning (assuming that the user did not press 'cancel').

guusdk:
Also (I have not tried this), have you considered a user:

- Adding the same certificate twice, using the same alias.
- Adding the same certificate twice, using a different alias.
- Adding two different certificates, but using the same alias.

What would happen in those cases? What _should_ happen?

Alameyo:
- Adding new certificate with same alias will overwrite old one. I can add dialog asking if user is sure to overwrite. 

- Adding twice same certificate with different aliases for now cause no other issue than having same certificate with different alias.

- Thing is that certificate dialog on your screen look different from what I see on mine (I don't see labels on the left on your screen for example), but I will try to fix that.

core/src/main/java/org/jivesoftware/spark/ui/login/CertificatesManagerSettingsPanel.java
Alameyo: You can attach only one Truststore and it is somehow direct representation of it so that shouldn't be a problem but real reason why I changed it into static is I wanted to do this:
`public void showCertificate() {
		CertificateDialog certDialog = new CertificateDialog(localPreferences,
				certificates.get(CertificatesManagerSettingsPanel.getCertTable().getSelectedRow()));
	}`

core/src/main/java/org/jivesoftware/spark/ui/login/CertificateDialog.java
Alameyo: Shame...

guusdk:
I don't think that overwriting is acceptable  - at least not without a very clear warning. It will lead to very unexpected behavior otherwise. An alternative here might be to generate aliases automatically. If you do that, you can ensure that they're unique. Also, you then don't need to show them to the user at all.

It should not be possible to have two times the same certificate (with a different alias) either. Again, this will lead to a lot of confusion. People that remove one of the duplicates will think that the certificate now no longer is used - while in reality it still is.

Note that Spark is used by a *lot* of people, and that you're more of an expert on certificates than any of them. Even if things are obvious to you, you should always wonder if the average user (for example, think of your parents) can use the UI without your help.

Alameyo:
I tested it more and the most of the certificates are shown perfectly fine. 
obraz
While some are missing labels as on your screen and here:
obraz

It might happen even if it is second copy of same certificate in Keystore (like on my screens) so reason isn't in different cert's data. I will look what cause it but now I have no idea.

core/src/main/java/org/jivesoftware/spark/ui/login/CertificatesManagerSettingsPanel.java
Alameyo: Now I left it as is but if you still feel that I should change it I will come back to this.

Alameyo:
Now certificate dialog show message on top when you add certificate. Depending on user input there are 3 new dialogs which inform in user tries to use the same alias as is already used in Truststore, the same certificate as exist already in store or alias is empty name. I changed all TextFields in certificate dialog to TextAreas which wrap lines at the end. I changed a bit buttons/radio buttons but I don't have problem with not displaying "Distrust" so you have to check if problem still exist. Default now returns RunetimeException. I blocked resizing of dialog as it sometimes mess with placement of components. Also I divided some of the code into functions to make it more clear.

guusdk:
Now we're getting somewhere! Here's a screenshot of your latest code, as how it showed for me:

image

There's a weird empty space on top - can you fix that? Also, the certificate wasn't scrolled to the top - that's odd too.

I was able to press "ok" without making a selection whether to trust or distrust the certificate.

When trying to test the error message for having no alias, I pressed "ok" without adding an alias. That, to my surprise, showed me "you already have this certificate" (which was true). Then I got back to the screen where I had to fill in an alias. Even when filling out an alias, I still got there. That's silly. When the user is adding a certificate that's already in the store, we can know this on the very first screen. There's no point in having the user to fill in an alias at all - it's going to fail anyway.

Alameyo:
> There's a weird empty space on top - can you fix that? Also, the certificate wasn't scrolled to the top - that's odd too.

I will try to fix that spacing at top , but that's problematic for me as I don't see it. This is how it appear for me.
obraz

> I was able to press "ok" without making a selection whether to trust or distrust the certificate.

I think I can setup it so it will be by default set as trust when adding certificate (why someone would like add certificate that he distrust?) unless it's invalid. User will be able to change it to distrust during adding or at any point later.

> When trying to test the error message for having no alias, I pressed "ok" without adding an alias. That, to my surprise, showed me "you already have this certificate" (which was true). Then I got back to the screen where I had to fill in an alias. Even when filling out an alias, I still got there. That's silly. When the user is adding a certificate that's already in the store, we can know this on the very first screen. There's no point in having the user to fill in an alias at all - it's going to fail anyway.

Ok, I will fix that.

wrooot:
You might tey to both use same skin. Also, it might look differently on different OS i suppose. Although that should be cross platform :)

Alameyo:
- I use now subject's CN - common name for certificate alias. If the certificate with the same common name already exist in Truststore then I add number at the end.

- ScrollPanel now is scrolled to the top.

-  I changed weighty of the infoLabel to 0.0 so that might help with that spacing but now I cannot check it myself.

- 
 Also that window showing certificate now is only window before adding where you have to confirm it. After that there is dialog that popup when certificate has been added and there is also dialog that shows up when certificate already exist within Truststore.

guusdk:
This looks good! The certificate is scrolled all the way to the top, and the weird empty space above the certificate is gone.

image

- I see that a new certificate is "distrusted" by default. I think you wanted "trust" by default?
- How do I delete a certificate?
- The list of all certificates still shows the "alias". That now has no value for the user, so we can remove that column, I think.
- Can we make the last few columns smaller? That would make the first few wider, and show more useful data, instead of empty space.

image

guusdk:
Oh, and I'd make the 'subject' column the first column, as that's more important than the issuer. Perhaps we shouldn't even show the issuer at all. That will only confuse people. What do you think?

Alameyo:
>     I see that a new certificate is "distrusted" by default. I think you wanted "trust" by default?
I decided I will make it trusted/distrusted depending on it's validity (which I still need to implement). If user still want to trust it he can change it by clicking trust/adding it to the exception list. End user might doesn't know that his certificate is invalid, this way we will make him sure that it is. I might add dialog that will say "This certificate is invalid, if you still want to use it check "trust"."

>     How do I delete a certificate?
I am going to implement it with next PR, here is [issue](
https://issues.igniterealtime.org/browse/SPARK-1945) for that.

>    The list of all certificates still shows the "alias". That now has no value for the user, so we can remove that column, I think.

Maybe I should add it at least in certificate dialog?  

>   Can we make the last few columns smaller? That would make the first few wider, and show more useful data, instead of empty space.

Ok, I will try to do it.
> Oh, and I'd make the 'subject' column the first column, as that's more important than the issuer. Perhaps we shouldn't even show the issuer at all. That will only confuse people. What do you think?

I have no strong opinion which fields are the most important to show here, so if you think that I should remove issuer then I will do it.

guusdk:
>> I see that a new certificate is "distrusted" by default. I think you wanted "trust" by default?

> I decided I will make it trusted/distrusted depending on it's validity (which I still need to implement). If user still want to trust it he can change it by clicking trust/adding it to the exception list. End user might doesn't know that his certificate is invalid, this way we will make him sure that it is. I might add dialog that will say "This certificate is invalid, if you still want to use it check "trust"."

I like that idea!

>> How do I delete a certificate?

> I am going to implement it with next PR, here is issue for that.

Ok!

> I have no strong opinion which fields are the most important to show here, so if you think that I should remove issuer then I will do it

If you don't have a strong opinion, that lets drop it from the table, and try to show the 'subject' better. Less is more! :)

guusdk:
Oh - and pressing the "Show Certificate" button gives a nullpointer when you don't have a certificate selected. Easy fix: don't allow the button to be pressed when there is no selection in the table.

wrooot:
Those buttons doesn't have to be so wide. This looks weird :) "Show certificate", "Choose file".

Alameyo:
@wrooot For me it is wide but not as wide as on Guus's screenshot.

wrooot:
I don't have this PR branch merged, so can't test how it looks now, but in the previous PR maybe window was a bit narrower and buttons looked ok. It seems that the window is stretched now and buttons gridbags have been stretched along.

Alameyo:
So that's how it look like now for me:
obraz

speedy01:
looks nice.  are the columns sort-able?  that would be nice if they could be.  That could save someone a lot of time having to look for exceptions or for a cert to make a change to.

Alameyo:
This columns are sortable but currently after sorting there is displayed wrong certificate if you click on them.

guusdk:
There's something weird with the table. Its content does not use all space on the right side. Could you have a look at that?

image

Also, when a certificate has a "CN" value, I think we can show only that. So, instead of a list like this:

- CN=Microsec e-Szigno Root CA,OU=e-Szigno CA,O=Microsec Ltd.,L=Budapest,C=HU
- CN=CNNIC ROOT,O=CNNIC,C=CN
- CN=Starfield Services Root Certificate Authority - G2,O=Starfield Technologies\, Inc.,L=Scottsdale,ST=Arizona,C=US

We'd see a list like this:
- Microsec e-Szigno Root CA
- CNNIC ROOT
- Starfield Services Root Certificate Authority - G2

That's a lot easier on the eyes. Obviously, when you open th certificate, you should still see everything. Also, certificates are not required to have a Common Name. In such case, let's still show everything.

Finally, in the 'valid' column, let's add a a description of why something is not valid (possibly in a different color - red or orange). Possible values:

- valid (in normal color)
- expired (in red, if the 'not after' date does not match)
- not valid yet (in red, if the 'not before' date does not match)
- ... others?

Oh, and could you center those values? That also looks a lot nicer, I think.

wrooot:
Guus are you using some unusual wide resolution?

guusdk:
Nope - but I'm using a non-Windows desktop manager, so that might account for some of the weirdness. This is actually good - Swing does allow the developer to prevent this, when applied properly. It's somewhat of a hassle to get it right, but we're better of seeing these issues now, than in production. :)

wrooot:
For me on Windows it opens without this gap on the right. But I'm not using this PR. Last one merged.

Alameyo:
I hope now it stretching good (this took me to long). Validity cells are green if certificate is valid, red if invalid, grey if self signed (in some situations, for end certificates it's not fine, on the other hand root CA are self signed and it is fine so I give to it distinguished color). Validity column values are centered. I left Subject column as is because length of it's value can differ and when centered it is almost unreadable.

I also added some basic validation (by date and self signed) in CertificateModel and I done some changes in i18n values.

guusdk:
This is looking great! I have a couple of small remarks, but I'll raise them as JIRA issues. That will allow us to merge this PR.

Issues:
- https://issues.igniterealtime.org/browse/SPARK-1951 Cell colors are wrong after sorting.
- https://issues.igniterealtime.org/browse/SPARK-1952 Don't mix "validity" with "self-signed" status.
- https://issues.igniterealtime.org/browse/SPARK-1953 Certificate details are editable.

PR #347, opened 2017-06-22
Minor tweaks

guusdk:
Lots of minor tweaks. @wrooot I'll leave it up to you if any of these are JIRA-issue-worthy.


wrooot:
At least on Windows some of the dialogs are always opening on top of a roster window and it is not saving the changed position. As i have my roster window docked in the right lower corner of the screen, the room browser opens in the corner also. But as you have made it higher, the Close button now goes beyond the lower edge of my screen. That's the ticket about dialogs positions https://issues.igniterealtime.org/browse/SPARK-945

PR #349, opened 2017-06-22
SPARK-945: Storing size and location of various screens.

guusdk:


wrooot:
Noticed a regression. Preferences window now has smaller default size (many settings don't fit now). And it doesn't save it's size after resizing it. It only saves its position. https://issues.igniterealtime.org/browse/SPARK-1956

PR #351, opened 2017-06-27
SPARK-1934, SPARK-1939 Certificate extensions.

Alameyo:


Alameyo:
I added new class translating OIDs. I could use just resources but i think that give some place to expand it in future with things like getFatherOID(), getChildredn() if I would need that. I couldn't extract some extensions. In this case such extension goes to TextArea unsupported extensions where it's OID and if possible it's description goes but not value. I think it's quite impossible to get all extensions/it's values as some can be custom in use by only particular companies (like Microsoft).

Also sometimes is shown here error with missing labels on left but it was resolved in other PR so to prevent conflicts I don't solve such things again.

guusdk:
I cannot open the "advanced" window any more. When I click on the button (in the login screen), this exception is being logged:

```
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
	at org.jivesoftware.sparkimpl.certificates.CertificateModel.extensionExtractHandler(CertificateModel.java:291)
	at org.jivesoftware.sparkimpl.certificates.CertificateModel.setupExtensions(CertificateModel.java:189)
	at org.jivesoftware.sparkimpl.certificates.CertificateModel.<init>(CertificateModel.java:172)
	at org.jivesoftware.sparkimpl.certificates.CertificateModel.<init>(CertificateModel.java:133)
	at org.jivesoftware.sparkimpl.certificates.CertificateController.<init>(CertificateController.java:88)
	at org.jivesoftware.spark.ui.login.CertificatesManagerSettingsPanel.<init>(CertificatesManagerSettingsPanel.java:74)
	at org.jivesoftware.spark.ui.login.LoginSettingDialog.<init>(LoginSettingDialog.java:58)
	at org.jivesoftware.LoginDialog$LoginPanel.actionPerformed(LoginDialog.java:737)
```

Alameyo:
Ok, seems I missed some situations when some values in extensions can be null. I will add some more checks for null values and I will add catching NullPointerException as it seems I cannot predict every possible situations for every extension.

Alameyo:
Now I also added here SPARK-1955 which is information about criticality of extensions.

core/src/main/java/org/jivesoftware/sparkimpl/certificates/CertificateModel.java
```diff
@@ -39,7 +62,24 @@
...
+	private String keyUsageExtension; // OID 2.5.29.15
+	private String subjectAlternativeNameExtension; // OID 2.5.29.17
+	private String issuerAlternativeNameExtension; // OID 2.5.29.18
+	private String basicConstraintsExtension; // OID 2.5.29.19
+	private String nameConstraintsExtension; // OID 2.5.29.30
+	private String CRLDistributionPointsExtension; // OID 2.5.29.31
+	private String policyMappingsExtension; // OID 2.5.29.33
+	private String authorityKeyIdentifierExtension; // OID 2.5.29.35
+	private String policyConstraintsExtension; // OID 2.5.29.36
+	private String extendedKeyUsageExtension; // OID 2.5.29.37
```
guusdk: You are keeping a lot of duplicate data in each instance of the Certificate Model. It's not _wrong_ but you will see that this quickly results in weird bugs (where one "duplicate" field has different content as compared to a field that should have the same values, but does not). I advice that you don't store duplicates at all.

core/src/main/java/org/jivesoftware/sparkimpl/certificates/CertificateModel.java
```diff
@@ -129,8 +169,227 @@ public CertificateModel(X509Certificate certificate) {
...
+
+	/**
+	 * Get values of the extension and format them into readable Strings.
+	 * @param cert
+	 * @param oid
+	 */
+	private void extensionExtractHandler(X509Certificate cert, String oid, boolean critical) {
+		try {
+			ASN1Primitive primitive;
+			if (oid.equals("2.5.29.9")) {
```
guusdk: Try to use constants, instead of string values for OIDs. Bouncy Castle has nice ones:

    final Extension result = new JcaX509CertificateHolder( certificate ).getExtensions()
                .getExtension( Extension.authorityKeyIdentifier );

core/src/main/java/org/jivesoftware/sparkimpl/certificates/CertificateModel.java
```diff
@@ -129,8 +169,227 @@ public CertificateModel(X509Certificate certificate) {
...
+				try {
+					primitive = JcaX509ExtensionUtils.parseExtensionValue(cert.getExtensionValue(oid));
+					SubjectDirectoryAttributes sub = SubjectDirectoryAttributes.getInstance(primitive);
+					subjectDirectoryAttributesExtension = Res.getString("cert.is.critical") + critical + "\n";
```
guusdk: Here, you're composing the text that the end user sees, correct? Should you do that in the Model (instead of the View)?

guusdk: I'm not sure: perhaps you need a "certificate-extension" view?

core/src/main/java/org/jivesoftware/sparkimpl/certificates/CertificateModel.java
```diff
@@ -129,8 +169,227 @@ public CertificateModel(X509Certificate certificate) {
...
+				} catch (CertificateParsingException e) {
+					Log.warning("Couldn't parse extended key usage extension", e);
+				}
+			} else {
+				addToUnsupported(critical, oid);
+			}
+		} catch (NullPointerException e) {
+			Log.error("Couldn't extract " + oid + ": " + OIDTranslator.getDescription(oid) + "extension.", e);
+			addToUnsupported(critical, oid);
+		}
```
guusdk: This is getting a very, very long method, which will be hard to maintain. Can we improve on the code style here?

core/src/main/java/org/jivesoftware/sparkimpl/certificates/CertificateModel.java
Alameyo: This CertificateModel is made mostly to prepare text to show in view windows. Model of certificate for other use is mostly covered by X509Certificate class which I might to extend in future but I don't feel that I need that now.

Alameyo:
I also deleted one constructor, that with pure String values as I don't think I will need it.

guusdk:
That's certainly an improvement to the code, thanks! The scroll issue, and the issue of the UI becoming very very messy remains. Can you improve on that, please?

guusdk:
I'm merging this, after @Alameyo took a blood-oath promising that the remaining issues, captured in  SPARK-1958, would be resolved later.

PR #352, opened 2017-07-01
SPARK-1945, SPARK-1957, SPARK-1936

Alameyo:
Deleting certificates, additional field with information about validity and fixed bug with not displayable information about expired/not valid yet certificate.

### DIFF
--- a/core/src/main/java/org/jivesoftware/spark/ui/login/CertificateDialog.java
+++ b/core/src/main/java/org/jivesoftware/spark/ui/login/CertificateDialog.java
@@ -14,11 +14,16 @@ import java.awt.Insets;
 import java.awt.Toolkit;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
+import java.io.IOException;
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+import java.security.cert.CertificateException;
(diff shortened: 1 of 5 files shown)


Alameyo:
Issues SPARK-1957 and SPARK-1936 are basically the same sorry :(

wrooot:
Mark one as a duplicate of another one and close as Duplicate.

Alameyo:
Ok done.

Alameyo:
>  Also, we now get three different ways that prevent a certificate from working:
> 
>    - in the overview table, you can select the 'exempt' checkbox
>    - in the dialog, you can toggle the 'distrust' radio button
>   -  you can delete the certificate completely
>
> So many options are confusing. Can we make the user experience better here?

I am thinking that my original idea indeed might be a bit too complicated for user. I am thinking that I should  kick out trust/distrust buttons and left only adding certificate to exceptions. Accept all valid certificate by default, don't accept invalid unless exempted. If someone want to distrust valid certificate then I would left for him only deleting it.

guusdk:
That would be an improvement, I think.

guusdk:
@Alameyo asked me to merge this, to avoid conflicts with other PRs. The remaining work will be done as part of SPARK-1941.

PR #353, opened 2017-07-07
SPARK-1941, SPARK-1954

Alameyo:
Added 3 new KeyStores. Added method for managing certificates between them.

The question from me is that: I am not sure if I can add this Keystores in way I did it or not. In core/src/main/java/org/jivesoftware/Spark.java I was looking for some way to install it but I am not sure if I should do it this and if I should how to do it? 

Other than that it is working.


guusdk:
I don't see a problem with what you're doing. You're basically creating them in a new directory named 'security' which has a location that is relative to the users's home directory. That makes sense to me. There's an added benefit that different users would have their own keystores.

Two things to improve:

- where possible, avoid using a String to represent something that has it's own implementation in Java (such as directories and files)
- use try-with-resources, as that's more fault-tolerant than the regular try/catch/finally.

Alameyo:
I have done this changes as well as some changes to make it work well with previous PR after rebasing.

core/src/main/java/org/jivesoftware/sparkimpl/certificates/CertificateController.java
```diff
@@ -100,14 +98,57 @@ public CertificateController(LocalPreferences localPreferences) {
...
+	}
+	
+	/**
+	 * Return information if certificate is on exception list.
+	 * @param alias
+	 * @return
+	 */
+	public boolean isOnExceptionList(String alias){
+		try (FileInputStream input = new FileInputStream(EXCEPTIONS)) {
+			trustStore = KeyStore.getInstance("JKS");
```
guusdk: This loads the entire keystore file, each time. That's probably very slow. You might want to keep it in memory.

core/src/main/java/org/jivesoftware/sparkimpl/certificates/CertificateController.java
```diff
@@ -127,20 +168,48 @@ private void fillCertTableWithKeyStoreContent(String storePath) {
...
 		}
 
 	}
-	
+/**
+ * Return file which contains certificate with given alias;
+ * @param alias
+ * @return
+ */
+	public File getAliasKeyStore(String alias) {
```
guusdk: Using the alias to identify a certificate is not a good idea, as the alias is something that anyone can set, and is not a unique value. Certificates themselves have unique values - you should use those!

core/src/main/java/org/jivesoftware/sparkimpl/certificates/CertificateController.java
Alameyo: Ok.

core/src/main/java/org/jivesoftware/sparkimpl/certificates/CertificateController.java
Alameyo: To some extend I need to use alias as deleteEntry/setCertificateEntry methods of the KeyStore work only with certificate alias. I am not sure if changing it for signature would help a lot, especially as it is now in use only for deleting certificates and to firstly get certificates I already had to do  

`X509Certificate certificate = (X509Certificate) keyStore.getCertificate(alias);`

Still I remember that for sure using unique values is important to identify certificates for creating SSLContext/TrustManager.

core/src/main/java/org/jivesoftware/sparkimpl/certificates/CertificateController.java
```diff
@@ -128,29 +134,15 @@ public void addOrRemoveFromExceptionList(boolean checked){
...
-				}
-			}
-			} catch (IOException | KeyStoreException | NoSuchAlgorithmException | CertificateException e) {
-				Log.error("Cannot access Keystore", e);
-			}
-		
-			return false;
-	}
-	
+	public boolean isOnExceptionList(CertificateModel cert) {
```
guusdk: You can simplify this a bit. :)

``return exemptedCertificates.contains(cert);``

Alameyo: Sure.

PR #356, opened 2017-07-24
SPARK-1966, SPARK-1952, SPARK-1969, SPARK-1970, SPARK-1971 and partially SPARK-1968

Alameyo:
Generally trust managers, checking CRL and some minor things. Still TO-DO: accepting self signed certificates, make use of blacklist keystore and do OCSP.


akrherz:
@Alameyo It would be good for your commit message to actually have the SPARK-???? nomenclature for each issue so that the Jira auto-linking works / this commit gets associated with the proper Jira issue.

Alameyo:
@akrherz Ok, title was generated from commit comment and I forgot to add SPARK

Alameyo:
Now also SPARK-1975 as Self Signed certificates can be accepted.

guusdk:
That last commit did not have SPARK-1975 in it either. :)

You can (and should!) rewrite those commit messages. There are a couple of options, see: https://help.github.com/articles/changing-a-commit-message/

core/src/main/java/org/jivesoftware/sparkimpl/certificates/SparkTrustManager.java
```diff
@@ -0,0 +1,387 @@
...
+            exceptionsTrustManager.checkServerTrusted(chain, authType);
+        } catch (CertificateException ex) {
+            // in case when certificate isn't on exceptions list then make use of this Trust Manager
+
+            // validate chain by date (expired/not valid yet)
+            checkDateValidity(chain);
+
+            // check if certificate isn't self signed, self signed certificate still have to be in TrustStore to be
+            // accepted
+            if (chain.length > 1) {
```
guusdk: Chain length is not a safe way to determine if the certificate in the chain is self-signed. The chain could be incomplete (it might depend on certificates that are in the truststore, without including those certificates), and chains might contain certificates that are not part of the chain at all (and are unused). You should find a better way to identify "self-signed"

Alameyo:
Added support for SPARK-1975 OCSP.

core/src/main/java/org/jivesoftware/sparkimpl/certificates/SparkTrustManager.java
```diff
@@ -187,6 +188,17 @@ public void checkServerTrusted(X509Certificate[] chain, String authType) throws
     }
 
     /**
+     * Return true if the certificate chain contain only one Self Signed certificate
+     */
+    private boolean isSelfSigned(X509Certificate[] chain) {
+        if(chain[0].getIssuerX500Principal().getName().equals(chain[0].getSubjectX500Principal().getName())){
```
guusdk: Perhaps replace by: `return chain[0].getIssuerX500Principal().getName().equals(chain[0].getSubjectX500Principal().getName())` (there's no need for the if/else)?

core/src/main/java/org/jivesoftware/sparkimpl/certificates/SparkTrustManager.java
```diff
@@ -187,6 +188,17 @@ public void checkServerTrusted(X509Certificate[] chain, String authType) throws
     }
 
     /**
+     * Return true if the certificate chain contain only one Self Signed certificate
+     */
+    private boolean isSelfSigned(X509Certificate[] chain) {
```
guusdk: This code assumes that the first certificate is the self-signed one, which could lead to several problems. For example, a certificate chain should, but is not always ordered (and if it is, I think it's typically the *last* certificate that's self-signed). 

Perhaps you should consider changing the argument to this method: instead of passing a chain, you might want to pass a single certificate. That way, this method's responsibility is very well defined. If needed, you could add additional methods that check if the entire chain represents a self-signed certificate (by checking if the chain size is equal to 1, and by passing that certificate to this method).

core/src/main/java/org/jivesoftware/sparkimpl/certificates/SparkTrustManager.java
```diff
@@ -189,24 +185,32 @@ private void validatePath(X509Certificate[] chain)
...
-            if (checkCRL) {
-                // check add CRL store and download CRL list to this store.
+            // check OCSP, CRL serve as backup
+            if (checkOCSP && checkCRL) {
+                // OCSP checking is done according to Java PKI Programmer's Guide, PKIXRevocationChecker was added in Java 8:
+                // https://docs.oracle.com/javase/8/docs/technotes/guides/security/certpath/CertPathProgGuide.html#PKIXRevocationChecker
+                PKIXRevocationChecker checker = (PKIXRevocationChecker) certPathBuilder.getRevocationChecker();
+                // soft fail option mean that OCSP or CRL must pass validation, in case when OCSP cannot be
+                // validated it will validate CRL and then if both cannot be checked it will fail
+                checker.setOptions(EnumSet.of(PKIXRevocationChecker.Option.SOFT_FAIL));
```
guusdk: Should this be configurable? Perhaps the end-user should control if SOFT_FAIL should apply.

core/src/main/java/org/jivesoftware/sparkimpl/certificates/SparkTrustManager.java
Alameyo: I added check if chain length equals 1. If it is true together as well as same subject issuer requirement then certificate is self signed. This way in case when chain would be reversed there will be no problems about that. 
One note: I don't worry here about standard root CA at the end of certificate path but about situation when server send only one self signed certificate for it's identification. If chain is longer than 1 then it is standard path building.

core/src/main/java/org/jivesoftware/AccountCreationWizard.java
```diff
@@ -353,7 +360,23 @@ private XMPPConnection getConnection() throws SmackException, IOException, XMPPE
         {
             builder.setHost( localPreferences.getXmppHost() );
         }
-
+        
+        if (securityMode != ConnectionConfiguration.SecurityMode.disabled && !useOldSSL) {
+            builder.setPort(5222);
```
guusdk: Does this take into account that the user can also set another port than 5222? If that's done, we should not hard-code to 5222.

core/src/main/java/org/jivesoftware/LoginDialog.java
```diff
@@ -277,6 +284,22 @@ protected XMPPTCPConnectionConfiguration retrieveConnectionConfiguration() {
             builder.setProxyInfo( proxyInfo );
         }
 
+        if (securityMode != ConnectionConfiguration.SecurityMode.disabled && !useOldSSL) {
```
guusdk: (same as before - can you get rid of this code duplication?) Does this take into account that the user can also set another port than 5222? If that's done, we should not hard-code to 5222.

core/src/main/java/org/jivesoftware/spark/ui/login/CertificatesManagerSettingsPanel.java
```diff
@@ -112,19 +114,30 @@ public Component prepareRenderer(TableCellRenderer renderer, int rowIndex,
...
 		ResourceUtils.resButton(acceptExpired, Res.getString("checkbox.accept.expired"));
-		ResourceUtils.resButton(acceptRevoked, Res.getString("checkbox.accept.invalid"));
+		ResourceUtils.resButton(acceptRevoked, Res.getString("checkbox.accept.revoked"));
 		ResourceUtils.resButton(acceptSelfSigned, Res.getString("checkbox.accept.self.signed"));
 		ResourceUtils.resButton(checkCRL, Res.getString("checkbox.check.crl"));
 		ResourceUtils.resButton(checkOCSP, Res.getString("checkbox.check.ocsp"));
 		ResourceUtils.resButton(showCert, Res.getString("button.show.certificate"));
 		ResourceUtils.resButton(fileButton, Res.getString("label.choose.file"));
-
+
```
guusdk: Small bug: when I open the "certificates"  tab and the "accept all"  option is already checked, then the others are not disabled. It works when I unselect and reselect again.

core/src/main/java/org/jivesoftware/AccountCreationWizard.java
Alameyo: Changed. I digged into that a bit more and while 5222 is recommended because other ports might be blocked by firewalls, there is no general prohibition against using other ports.

core/src/main/java/org/jivesoftware/spark/ui/login/CertificatesManagerSettingsPanel.java
Alameyo: Fixed.

core/src/main/java/org/jivesoftware/LoginDialog.java
Alameyo: Indeed this two classes share a bunch of the similar code, tomorrow I will try to reduce some of its redundancy.

PR #358, opened 2017-07-26
SPARK-1972: New 'security' login settings panel.

guusdk:
spark-security-login-panel
This commit adds a new 'security' tab to the advanced login settings. Here, security-related
settings are made available.

Also, this commit makes slight modifications to the other tabs, by making the components be
oriented more to the top-left corner (instead of being centered - which looks pretty ugly).

### DIFF
--- a/core/src/main/java/org/jivesoftware/LoginDialog.java
+++ b/core/src/main/java/org/jivesoftware/LoginDialog.java
@@ -30,6 +30,7 @@ import org.jivesoftware.smack.sasl.javax.SASLExternalMechanism;
 import org.jivesoftware.smack.sasl.javax.SASLGSSAPIMechanism;
 import org.jivesoftware.smack.tcp.XMPPTCPConnection;
 import org.jivesoftware.smack.tcp.XMPPTCPConnectionConfiguration;
+import org.jivesoftware.smack.util.DNSUtil;
 import org.jivesoftware.smack.util.TLSUtils;
 import org.jivesoftware.smackx.chatstates.ChatStateManager;
 import org.jivesoftware.spark.SessionManager;
@@ -211,7 +212,8 @@ public class LoginDialog {
             }
         }
 
-        boolean useSSL = localPref.isSSL();
+        ConnectionConfiguration.SecurityMode securityMode = localPref.getSecurityMode();
+        boolean useOldSSL = localPref.isSSL();
         boolean hostPortConfigured = localPref.isHostAndPortConfigured();
 
         ProxyInfo proxyInfo = null;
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/jivesoftware/LoginDialog.java
```diff
@@ -273,16 +275,22 @@ protected XMPPTCPConnectionConfiguration retrieveConnectionConfiguration() {
...
+                // Here, we force the host to be set (by doing a DNS lookup), and force the port to 5223 (which is the
+                // default 'old-style' SSL port).
+                builder.setHost( DNSUtil.resolveXMPPDomain( loginServer, null ).get( 0 ).getFQDN() );
                 builder.setPort( 5223 );
             }
             builder.setSocketFactory( new DummySSLSocketFactory() );
+            // SMACK 4.1.9  does not recognize an 'old-style' SSL socket as being secure, which will cause a failure when
+            // the 'required' Security Mode is defined. Here, we work around this by replacing that security mode with an
+            // 'if-possible' setting.
+            builder.setSecurityMode( ConnectionConfiguration.SecurityMode.ifpossible );
```
Alameyo: Isn't that misleading setting SecurityMode.ifpossible while there could be already set up other mode? Unless old SSL require it I think it should be changed to builder.setSecurityMode( securityMode );

guusdk: It's not an ideal solution, I agree. But, when using 'old-style SSL', we cannot set security mode "required" (which is what you'd ideally want, I think). 

We cannot do this, because Smack (at least in version 4.1.9) will not detect that the connection is encrypted when using 'old-style SSL'. If you set the mode to 'required', Smack will throw an error (as it didn't detect encryption, but knows that encryption is required).

guusdk: Perhaps it would be a bit nicer to do what I did there only when the securityMode was set to 'required':
`if ( securityMode == required) { builder.setSecurityMode( ifpossible ); }`
But as there are only three enum values (required, ifpossible, disabled), the behavior would be the same.

Alameyo: If oldSSL need ifpossible mode that then it is ok.

guusdk:
@wrooot I'm merging this without your review, as I'd like @Alameyo to be able to rebase his GSoC work onto. Feel free to raise concerns though - I'll address them later.

wrooot:
It works. But it shows a generic Unknown error message and one can only say what happened by checking the logs "WARNING: Exception in Login:
org.jivesoftware.smack.SmackException$SecurityRequiredByClientException: SSL/TLS required by client but not supported by server". Would be good to have a proper message in Spark.

guusdk:
You're right @wrooot. I have created https://issues.igniterealtime.org/browse/SPARK-1976 to track improving the error message.

PR #359, opened 2017-07-26
SPARK-1973: Spark should show an indicator when the connection is not…

guusdk:
… encrypted.


core/src/main/resources/i18n/spark_i18n.properties
```diff
@@ -853,6 +853,7 @@ message.server.unavailable = Can't connect to server: invalid name or server not
 message.service.already.exists = Service is already in your service list
 message.shared.group = Shared group
 message.spark.secure = Spark is running in secure mode
+message.spark.unsecure = Spark is not running in secure mode (the connection the server is not encrypted).
```
wrooot: Probably should be "to the server"

guusdk: You are obviously right. Feel free to fix errors like that yourself (I think you should be able to modify the github PR directly in the web console).

wrooot: Ok, i did

PR #360, opened 2017-07-27
SPARK-1976: Better error messages

guusdk:
This commit:
- reworks the login method to remove some redundant code;
- applies the MessageDialog-provided error dialog generator to the login method;
- refactored the MessageDialog error dialog generator (to allow for toggling of the large stacktrace pane).

In all, this should improve the error messages during login. The same technique could be re-used elsewhere.


guusdk:
This intends to address @wrooot's concern in #358

wrooot:
Is it possible to make the error dialog to be in the center of the screen or at least on top of the login window? It always starts a bit to the right. And the details window is too big, so in my case after opening it i can't reach buttons at the bottom without moving the window. Not a huge issue though.

guusdk:
I created a new PR for that. See #361

PR #362, opened 2017-07-28
SPARK-1888: Plugin management

guusdk:
Should fix the unloading of plugins (and thus, fix the re-installing and updating of plugins). Please note that this PR is completely untested by me - @wrooot - could you do the honors please?


wrooot:
Merged. Fixed the issue ;)

PR #363, opened 2017-07-28
SPARK-1956: Preference dialog should remember its size.

guusdk:
The recent fix for SPARK-945 already stored the preference dialog size and position, which fixed most of this bug. When that was implemented, a rogue invocation of .pack() was placed after the stored settings were applied. This resized the component again.


wrooot:
It now saves the size and position. But maybe you can make its initial size bigger, so settings on the Appearance tab would fit? Interesting though, that Roar settings uses scrollbars when it doesn't fit. All the other tabs do not show scrollbars.

PR #368, opened 2017-08-13
Openfire meetings pluginfor Spark 0.4.0

deleolajide:
This PR updates the spark meet plugin. I have done the following:

made meet plugin compatible with openfire ofmeet plugin 0.9.x
made meet plugin compatible with maven build
replaced external electron chromium web browser with built-in jxbrowser
upgraded POM to pull latest 6.14 jxbrowser artifacts
updated the expired evaluation license of jxbrowser. This replacement will expire on 6th of Sept 2017.

What is the plan for a proper OSS license for JxBrowser?

image2


wrooot:
Oh, i think Guus and speedy were working on a replacement for JxBrowser (as it was a complicated matter with the license)..

deleolajide:
Thats @wrooot I gave up on JxBrowser for my [Pade ](https://github.com/igniterealtime/Pade)project because of probably the same complications

wrooot:
Guus can tell more, but i think they had at least some progress in making JCEF to Maven or generating binaries or something along these lines.

deleolajide:
I gave up on JECF as well and used Electron instead. See my [igniterealtime blog](https://community.igniterealtime.org/blogs/ignite/2016/10/30/sparkmeet-spark-plugin-for-openfire-meetings)

speedy01:
> replaced external electron chromium web browser with built-in 

I'm curious why you removed the electron chromium and went back to jxbrowser? With unable to get a straight answer from jxbrowser on the OOS license, its likely we'll continue to look for an alternative.  jcef or something else.  Did you hit some road blocks with electron?

mariushamilton:
If you guys want me to try something let me know. 
I have over 150+ users in my organization

deleolajide:
>Did you hit some road blocks with electron?

Electron is awesome. There is a [desktop version of Jitsi Meet web application](https://github.com/jitsi/jitsi-meet-electron) used in Openfire Meetings built with Electron. It even has desktop remote control working alongside WebRTC screen share.

speedy01:
if electron is the better choice, would it make since to create 3 separate plugins (linux/mac/windows), and bundle the plugin based on os build, or offer the plugin as a separate download?

deleolajide:
There was a vote conducted by @wrooot about this on igniterealtime. If I remember correctly, most people were not in favour of increasing the already inflated size of Spark. 

It makes sense to reuse the embedded browser already included in Spark whatever it may be as long as it supports WebRtc and Jitsi Meet

wrooot:
Here's that poll. Actually, most people didn't care about the size increase and even longer startup https://community.igniterealtime.org/polls/1105

deleolajide:
>Actually, most people didn't care about the size increase
thanks for the correction @wrooot. 

>or offer the plugin as a separate download?
It might be better to move the meet plugin to Openfire Meetings (ofmeet) where we can build the separate platform versions and make them downloadable links. 

I like Electron, but there is still a bit work to install the platform dependent executable binaries in the right places for the JVM to find them. JxBrowser does all that and more and is already working fine with Spark. I vote to keep JxBrowser in short term and then plan for Electron in the long term.

guusdk:
@deleolajide - this is awesome! I've got a lot more to say, but a celebration of your effort on this deserves a comment on its own. :)

guusdk:
I've been talking to various people through different channels on this, but it will be good to centralize that discussion here. Forgive me for still being somewhat in holiday mode. :)

For starters: this is awesome! Having video-conference (back) in Spark is a great feature! @deleolajide 's work also build on what's now in the repository, so we shouldn't argue to much about what the long term solution should be in context of accepting or not accepting this PR (I intend to merge it after writing this). If our long-term plan is to move in any other direction, we should simply take into account migrating everything that depends on it - including Dele's work here.

@wrooot is likely concerned about releasing a version of Spark with JxBrowser included. We have not done that so far (JxBrowser is in Master only, of which to date, no release has been made). I don't think we should do this. As @speedy01 mentioned, the JxBrowser licensing that they've offered us was deemed to vague by various community members - not something to add to Spark as-is. If desired, we can discuss the offered license in detail, but lets do that outside of this PR (you know where to find me).

@deleolajide described an in-between solution in a one-on-one I had with him earlier this week: perhaps we could ship Spark with JxBrowser, but without a license at all - requiring end-users to obtain a license of their own before being able to make use of the functionality. We'd have to check if that's OK, from a licensing perspective (we'd still be distributing their binaries). This might be a good stopgap solution.

My preference for a long term goal would be to bring JCEF up to spec with JxBrowser, if we can't get a proper license for JxBrowser. Work on that [has started](http://magpcss.org/ceforum/viewtopic.php?f=17&t=15307&sid=7fc0c8ba872b85487b4d810bdf2d4667), but needs additional love. I will, however, note that I prefer JCEF to Electron simply because I'm familiar with JCEF - which is a weak argument.

deleolajide:
>Perhaps we should generate a different license?

I think we should remove it now and inform users to obtain and install their own license.jar file as you advised above unless you have a license with long life.

speedy01:
so..here is my 2 cents...for what its worth.  echoing what I said in open_chat

According to 18 USCS § 31, term "used for commercial purposes" means the carriage of persons or property for any fare, fee, rate, charge or other consideration, or directly or indirectly in connection with any business, or other undertaking intended for profit. 

IRT would be issued the license, and as long as IRT is not profiting from jxbrowser, than we should be ''ok". 

Since TeamDev doesn't have an OpenSource license, it looks like they would issue a free Commercial License.  Hopefully this would be issue to as a Project or Company license.  If project or company, than we should be covered https://www.teamdev.com/jxbrowser/licence-agreement#DevelopmentandRuntimeLicenceGrant on developers.  However, "...by any number of developers, employees or contractors under your control" verbiage bothers me. Mainly the "...under your control." 

The licensing is vague at best, which is scary and doesn't seem to opensource friendly.    I'm very grateful to @deleolajide for his work, and would love to see it sooner vs later. However,  I think jxbrowser should be used as a last resort. 

If we ask users to obtain their own license, I'm afraid no one will since it would creat additional complexity that many people may not be well equipped for.

wrooot:
As Guus suggested, adding a comment here. I have filed https://issues.igniterealtime.org/browse/SPARK-1987 with concerns about the installer being inflated because of binaries for all platforms included in a platform specific installer. In the past we (speedy and I) have worked to get rid of linux specific libraries in a Windows installer (and vice versa). So, this is a bit of a regression now.

PR #371, opened 2017-08-14
SPARK-1985 Support for TLS mutual authentication

Alameyo:
Few things to improve (like refreshing certificate table which I will try to add soon), but generally it is working. One thing is Spark can add key entries only from PEM file (without password) as this file can contain both certificate and private key. In future there might be implemented support for some other formats but I think it is enough for now.


core/src/main/java/org/jivesoftware/spark/ui/login/MutualAuthenticationSettingsPanel.java
```diff
@@ -0,0 +1,265 @@
...
+           addCertificate();
+        }
+        if(e.getSource() == idTable){
+            showCert.setEnabled(true);    
+        }
+        if(e.getSource() == showCert){
+            idControll.showCertificate();
+        }
+        if (e.getSource() == createButton) {
+            idControll.setUpData(commonNameField.getText(), organizationUnitField.getText(),
```
guusdk: Could you move this to a dedicated method?

core/src/main/java/org/jivesoftware/spark/ui/login/MutualAuthenticationSettingsPanel.java
```diff
@@ -0,0 +1,265 @@
...
+        fileChooser.setFileFilter(certFilter);
+
+        int retVal = fileChooser.showOpenDialog(this);
+        if (retVal == JFileChooser.APPROVE_OPTION) {
+
+            File file = fileChooser.getSelectedFile();
+            try {
+                idControll.addEntryToKeyStore(file);
+                // idControll.addCertificateToKeystore(file);
+            } catch (IOException e) {
```
guusdk: Instead of writing this to a log file (or perhaps, in addition to writing it to a log file), it would be good to make an error message pop up. Most users don't look at the log file, and will not know that something went wrong.

core/src/main/java/org/jivesoftware/sparkimpl/certificates/CertificateModel.java
```diff
@@ -362,7 +362,7 @@ private boolean isSelfSigned(){
 	}
 	
 	private boolean checkValidity() {
-		if (isAfterNotAfter() && isBeforeNotBefore() && checkRevoked()) {
+		if (!isAfterNotAfter() && !isBeforeNotBefore() && !checkRevoked()) {
```
guusdk: Please add javadoc to describe when true or false is expected to be returned.

core/src/main/java/org/jivesoftware/spark/ui/login/MutualAuthenticationSettingsPanel.java
Alameyo: I made 2 methods: one for creating Certificate and one for creating CSR.

PR #372, opened 2017-08-16
Disable meet plugin by default.

deleolajide:
This PR is to address SPARK-1987 by disabling the auto build of meet plugin. Not sure how the plugin inflates the installer as it uses a standard plugin pom file and includes Java classes from Spark Core. The change is to eliminate it from the issue.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -76,7 +76,7 @@
         <module>plugins/flashing</module>
         <module>plugins/growl</module>
         <!--<module>plugins/jingle</module>-->
-        <module>plugins/meet</module>
+        <!--<module>plugins/meet</module>-->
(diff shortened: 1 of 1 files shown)


wrooot:
Dele, thanks for the quick response. I don't know either why all the binaries were included. Maybe that's a Maven thing. As the first time SparkMeet has been included into the source it was still using Ant build. Or the thing that it is now bundled (it wasn't before) was the cause of this issue appearing now.

wrooot:
Strange. The size hasn't changed. Still around 240 MB. I'm referring to nightly builds https://bamboo.igniterealtime.org/browse/SPARK-NIGHTLY-228/artifact/shared/Install4j-generated-media/

PR #373, opened 2017-08-17
SPARK-1987 Update install4j to jx-browser ver 6.14

deleolajide:
The Spark install4j installer script is hard-coded to use JxBrowser version 6.12 and is unable to use the dependency version info from maven pom file. This PR upgrades it to use 6.14 which is required for Spark Meet plugin.

This should resolve SPARK-1987

### DIFF
--- a/distribution/src/installer/spark.install4j
+++ b/distribution/src/installer/spark.install4j
@@ -576,12 +576,12 @@
       <excludedBeans />
       <overriddenPrincipalLanguage id="en" customLocalizationFile="" />
       <exclude>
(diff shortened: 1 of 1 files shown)


wrooot:
Thanks, the size is back to normal now ;)

deleolajide:
>Thanks, the size is back to normal now ;)

Would like to re-enable the meet plugin and delete the outdated teamdev license file from the Spark core jar file as it will stop the plugin from working without a re-build and take precedence over an external license jar file being applied at install.

PR #376, opened 2017-08-19
SPARK-1989 create empty KeyStores

Alameyo:
New empty KeyStores are created each time when user open advanced panel and there are no existing KeyStores in user directory already.


wrooot:
What does it mean that user opens advanced panel? What if a user doesn't ever open it and just hits login on a fresh profile?

Alameyo:
Then KeyStores are opened if exist, if not new empty KeyStores are created because TrustManagers/KeyManager need them but I don't save them.

wrooot:
The usual user-friendly model is that if an unknown/self-signed certificate is encountered, the app provides a dialog/popup with an immediate option to add this certificate to an exception list or do a one time login. A number of clients i have used do this, browsers do this. It can involve a few steps, but a user shouldn't be looking for a menu to add a certificate to exceptions. A list of default certificates won't solve this.

Alameyo:
I agree, that is thing I should to add. I will work on that

guusdk:
The popup is certainly desirable, but not part of the GSoC project. Let's focus on finishing that first. We'll create the pop-up directly afterwards.

guusdk:
Thanks :)

PR #377, opened 2017-08-22
SPARK-1994 Make use of the JRE certificates.

Alameyo:


core/src/main/java/org/jivesoftware/sparkimpl/certificates/GeneralTrustManager.java
```diff
@@ -0,0 +1,85 @@
...
+                // Retrieve all of the certificates out of the keystore
+                // via the alias name.
+                int i = 0;
+                while (aliases.hasMoreElements()) {
+                    X509Certs[i] = (X509Certificate) allStore.getCertificate((String) aliases.nextElement());
+                    i++;
+                }
+            }
+        } catch (Exception e) {
+            Log.error(e.getMessage(), e);
```
guusdk: I don't really like `Log.error( e.getMessage(), e)`, as it'll lead to duplicate text in the error. Perhaps you should change `e.getMessage()` with a short text describing what the code failed to do.

core/src/main/java/org/jivesoftware/sparkimpl/certificates/CertificateController.java
```diff
@@ -41,17 +38,24 @@
  */
 
 public class CertificateController extends CertManager {
-	public final static File TRUSTED = new File(Spark.getSparkUserHome() + File.separator + "security" + File.separator + "truststore");
-	public final static File BLACKLIST = new File(Spark.getSparkUserHome() + File.separator + "security" + File.separator + "blacklist");
-	public final static File EXCEPTIONS = new File(Spark.getSparkUserHome() + File.separator + "security" + File.separator + "exceptions");
-	
-	
-	private KeyStore trustStore, blackListStore, exceptionsStore;
+	public final static File TRUSTED =            new File(Spark.getSparkUserHome() + File.separator + "security" + File.separator + "truststore");
```
guusdk: Please add in javadoc exactly what the purpose of every store is - so that Devs that look at this in the future understand how things are combined, and why.

core/src/main/java/org/jivesoftware/sparkimpl/certificates/GeneralTrustManager.java
Alameyo: Ok, I changed exception for KeyStoreException and added message.

core/src/main/java/org/jivesoftware/sparkimpl/certificates/CertificateController.java
Alameyo: I added nice comment describing all Files/KeyStores

PR #378, opened 2017-08-24
SPARK-1995, SPARK-1996, SPARK-1997 and lack of resources in one text.

Alameyo:
Refreshing certificate table works. Blacklist is now used for adding status to the Certificate Model, not as separate List in table. Also now button for checking revocation in certificate dialog works.

### DIFF
--- a/core/src/main/java/org/jivesoftware/spark/ui/login/MutualAuthenticationSettingsPanel.java
+++ b/core/src/main/java/org/jivesoftware/spark/ui/login/MutualAuthenticationSettingsPanel.java
@@ -266,7 +266,7 @@ public class MutualAuthenticationSettingsPanel extends JPanel implements ActionL
             } catch (InvalidKeySpecException | NoSuchAlgorithmException | KeyStoreException | InvalidNameException
                     | IOException e) {
 
-                JOptionPane.showMessageDialog(null, "dialog.cannot.upload.certificate");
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/jivesoftware/spark/ui/login/MutualAuthenticationSettingsPanel.java
```diff
@@ -266,7 +266,7 @@ private void addCertificate() {
             } catch (InvalidKeySpecException | NoSuchAlgorithmException | KeyStoreException | InvalidNameException
                     | IOException e) {
 
-                JOptionPane.showMessageDialog(null, "dialog.cannot.upload.certificate");
+                JOptionPane.showMessageDialog(null, Res.getString("dialog.cannot.upload.certificate"));
```
guusdk: This works as intended, but it introduces a small issue with the i18n file. The text contains quotes (which are displayed, which is ugly):

    dialog.cannot.upload.certificate = "Cannot upload certificate file"

should be

    dialog.cannot.upload.certificate = Cannot upload certificate file

It happens in a couple of places. Please fix :)

Alameyo: Fixed :D

PR #380, opened 2017-08-27
SPARK-2000 Default Settings and lacking resource for mutual.auth

Alameyo:
At first I was intending to add default settings only to Certificate Manager but as UseDefault button wasn't working for other tabs so I fixed them too. I added default settings to Default class and default.proporties and used it in tabs.


guusdk:
I'm confused - I could swear that I've seen the code that you're now adding in Spark before. I can't find it though. :)

PR #381, opened 2017-08-27
SPARK-1990 Remove PKI tab

Alameyo:
This PR removes PKI tab as with new Certificates Manager it no longer serves it's purpose.  It's not to instant merge as one thing I am not 100% sure is deleting things from LoginDialog as it seems this was also used for SASL EXTERNAL which seems to not work anyway, but maybe there is an easy fix I have no idea about yet.


guusdk:
Hmm, this now has conflicts. Could you rebase, please?

I'm concerned about two things:

1. We might want to have an option that enables/disables mutual authentication explicitly.
2. We should use SASL External (which is being removed in this PR).

Alameyo:
Ok, I will rebase.
1. Currently in Security tab there is "Allow for mutual authentication" checkbox.
2. Yes, I am also concerned about this. Maybe not merge it yet, I will try to make SASL External working.

PR #383, opened 2017-09-01
SPARK-2006 Don't use common name as alias if alias already exist.

Alameyo:
I probably introduced this bug with adding JRE certificates, but it is not related to random NullPointerExceptions. The problem which it generate is that not all certificates from JRE are displayed and that cause further problems at moving certificates to/from exceptions list.


guusdk:
Thanks!

PR #386, opened 2017-09-10
SPARK-2011 Bring back functionality for "Accept all" certificates

Alameyo:
Now again this will accept all certificates regardless of their validity or being a part of the TrustStores.

### DIFF
--- a/core/src/main/java/org/jivesoftware/sparkimpl/certificates/SparkTrustManager.java
+++ b/core/src/main/java/org/jivesoftware/sparkimpl/certificates/SparkTrustManager.java
@@ -59,7 +59,7 @@ import org.jivesoftware.spark.util.log.Log;
  */
 public class SparkTrustManager extends GeneralTrustManager implements X509TrustManager {
 
-    
+    private boolean acceptAll;
     private boolean checkCRL;
     private boolean checkOCSP;
     private boolean acceptExpired;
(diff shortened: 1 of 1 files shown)


guusdk:
Do we want to have this functionality? If we get a popup each time that we receive an untrusted certificate, then there is no much use for an "accept all" anymore.

guusdk:
Do we want to have this functionality? If we get a popup each time that we receive an untrusted certificate, then there is no much use for an "accept all" anymore.

Alameyo:
I think we can left this and still have popup. Just in case when this is ticked popup will not be shown, client will auto trust everything.

guusdk:
From a security perspective, I dislike a "trust all" option. Users can easily forget that it is enabled. The popup offers the same feature, but on a per-certificate basis. I like that a lot better.

wrooot:
Currently by default Accept all is disabled (in 2.8.3 fresh install). So, we can still have it, so it won't affect current installations and admins won't have to instruct their users suddenly (for a local server that's not a real issue).

wrooot:
Wait, but there is already Accept all in the current builds. What does it do then? If we going to not bring back Accept all, then it should be removed from the Certificates tab as well.

Alameyo:
Ok, I will remove that while adding popup.

PR #389, opened 2017-09-22
SPARK-1869 Add option to force In-band bytesteams for file transfers

speedy01:


wrooot:
Oh, i just now saw you requested a review :) Anyway, i will try to test this myself.

wrooot:
Have tested this out. Works fine. It is actually enough to have this enabled on the sender part and have older Spark on the other side to make it work. I haven't checked the speeds. It is fast enough to send a pic or a document (which it should be used for mostly). Thanks, speedy ;)

PR #391, opened 2017-09-27
SPARK-2001 Add popup for accepting certifcate from connection.

Alameyo:
Seems it works :) On the other hand I don't think I am able to remove that timeout popup.


Alameyo:
I added removal of accept all checkbox as it is kin of related.

guusdk:
Awesome! We'll work on that timeout popup later.

One thing that goes wrong: if the server has a self-signed certificate (that is in the store), you still get the question if you want to add it to the store.

I don't think we need two screens, by the way. Can't you combine them into one? "The server uses this certificate. Do you want to add it?"

Alameyo:
> One thing that goes wrong: if the server has a self-signed certificate (that is in the store), you still get the question if you want to add it to the store.

Solved, other things also were throwing CertificateException and I didn't tested it well so popup was showing up when anything was wrong. Now I show popup only when I catch CertPathValidatorException.

> I don't think we need two screens

Ok. Now one screen, with new custom message.

Also I check if certificate is invalid. In this case I add it to ExceptionsStore, not TrustStore.

One more thing, one line change: I noticed that Spark stopped accepting invalid exempted certificates. I remember I was testing it and it was working but it seems I didn't included removing Bouncy Castle provider from SparkExceptionsTrustMenager, which I did now. BC provider do some checks even when I set some properties to disable them, while default provider works fine.

PR #393, opened 2017-10-01
Spark-2017

speedy01:


wrooot:
We are not in a hurry to support Java 9. So will let Guus and alameyo to look at it and see if that will work with Java 8 ok.

speedy01:
oops..wrong button!  fyi..tested with java 8, and works as expected.

Alameyo:
If I understand correctly it is used for miniature Icons during file transfer? Seems it still works.  Also reason for this error is here [JEP-260](http://openjdk.java.net/jeps/260).

PR #394, opened 2017-10-05
SPARK-2001 Add popup for accepting certifcate from connection.

Alameyo:


wrooot:
I have merged this PR to see how it works. I agree with @guusdk older comments that there shouldn't be so many info and popups. Ideally it should only show a window with brief information about a cert (name, issuer maybe) with an option to see details. 

Also, there is no need to show a confirmation dialog after it has been added to a trust store. If possible it should just proceed with the login process automatically. And i would also remove the dialog telling that a cert was successfully deleted. This is already obvious as you see it disappearing from the list. Less dialogs for obvious effects is better.

No matter how fast i try to press OK and OK, it still shows an error: Certificate not in TrustStore (right after the dialog that my cert was added).

And something has to be done with that default timeout. User should be able to read about the cert and approve it without timeout errors (maybe this error can be suppressed from jumping forward and also from being logged).

wrooot:
IntelliJ reports these as not being used (so maybe they should be removed):

dialog.certificate.title.certificate.not.in.truststore
dialog.certificate.click.yes.if.you.want.to.see.certificate

wrooot:
IntelliJ reports these as not being used (so maybe they should be removed):

dialog.certificate.title.certificate.not.in.truststore
dialog.certificate.click.yes.if.you.want.to.see.certificate

wrooot:
IntelliJ reports these as not being used (so maybe they should be removed):

dialog.certificate.title.certificate.not.in.truststore
dialog.certificate.click.yes.if.you.want.to.see.certificate

Alameyo:
Ok. I will work on the next iteration of this. Well previous version had first screen asking to show certificate. No confirmation dialogs are doable. I am not sure yet exactly how to fix timeout notification but I will try to do something. 

This two dialogs were used when I had first screen asking to show certificate and I forgotten to delete them. I guess I will use them again then. 

> No matter how fast i try to press OK and OK, it still shows an error: Certificate not in TrustStore (right after the dialog that my cert was added).

Not sure about that, I don't think I see that

wrooot:
I was testing with my local test server with a self-signed cert.

wrooot:
See the last mentioned error on a fresh Spark profile also. Maybe you should try with a fresh profile and a self-signed certificate.

PR #398, opened 2017-10-23
Update README.md

AustinBrrtt:
The Spark README seemed a bit lacking in comparison to OpenFire's, so I copied the same format from OpenFire, but switched all the links to match Spark. I also edited the community info to mention Discourse instead of the old Community.


guusdk:
Much appreciated!

PR #399, opened 2017-11-07
Update ukrainian translations

jamazurak:
Translation updated to the newest version. Missing keys translated except OID's (as for now).
Some grammatical errors fixed (i.e. not_something is single word in most cases in Ukrainian; few other small errors i.e. incorrect letters in previous translation).
"Room" translated as room, not as "group" or "conference", because all those word are occured.
"Cancel" translated as cancel instead of "declined".
Key accelerators not touched, not added nor checked for duplicates.
Different translations for one thing should be fixed in the future; ukrainian translation conventions are welcome.

Translations is done in Eclipse by ResourceBundle Editor (http://essiembre.github.io/eclipse-rbe/).


wrooot:
Thanks for your contribution

PR #403, opened 2017-11-15
fix-logic-error-in-UserSearchResult-getFirstValue

dmitrii-88:


wrooot:
Thanks for you patch. I have assigned @guusdk to check it. Seems good, but as i'm not a dev myself, it is better to check :)

wrooot:
Thanks @Alameyo. Will merge it then.

dmitrii-88:
Because of this error person search does not work.

wrooot:
Could it also fix https://issues.igniterealtime.org/browse/SPARK-1873

dmitrii-88:
Yes, it could.

wrooot:
Will ask in the forums if someone can test it.

wrooot:
So far one confirmation that search works again.

PR #406, opened 2017-11-19
SPARK-1999 Fix Nullpointer that sometimes happens when user add/remove certificate from exceptions

Alameyo:
I was intending to reduce number of times when this error happens with this commit but it seems that I cannot reproduce error again after this changes. 

I reworked and simplified a bit moving certificates to/from exceptions. With this change exempted certificates are used as information to set "tick box" in certificate table instead of displaying separate list of certificates. 
That means I only read certificates from TrustStores while ExceptionsStores are used for management. 

I hope it solves problem with random NullPointerException because I can't reproduce it now (but I still don't know what caused it so it might be that still happen, just much less frequently).

### DIFF
--- a/core/src/main/java/org/jivesoftware/sparkimpl/certificates/CertificateController.java
+++ b/core/src/main/java/org/jivesoftware/sparkimpl/certificates/CertificateController.java
@@ -118,7 +118,7 @@ public class CertificateController extends CertManager {
                     String alias = (String) store.nextElement();
                     X509Certificate certificate = (X509Certificate) caStore.getCertificate(alias);
                     // if getCertificateAlias return null then entry doesn't exist in distrustedCaStore (Java's default).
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/jivesoftware/sparkimpl/certificates/CertificateController.java
```diff
@@ -206,67 +204,107 @@ private void addRowsToTableModel(List<CertificateModel> certList, Object[] certE
...
+        try {
+            if (exceptionsStore.getCertificateAlias(cert.getCertificate()) != null) {
+                return true;
+            } else if (exceptionsCaStore.getCertificateAlias(cert.getCertificate()) != null) {
+                return true;
+            } else {
+                return false;
+            }
+        } catch (KeyStoreException e) {
+            return false;
```
guusdk: If an error occurs, we shouldn't assume 'yes' or 'no'. Just throw or show the error instead.

PR #409, opened 2017-12-09
SPARK-2039: no encoded i18n files in source

guusdk:
Should we increase the version number of all plugins? This shouldn't be a functional change...


wrooot:
I think we shoouldn't change versions then.

PR #412, opened 2018-01-07
SPARK-2045 Make better use of PKIXRevocationChecker

Alameyo:
Use of PKIXRevocationChecker instead of custom code for downloading CRLs. This way checker's options are better used (and soft fail policy can be set with just CRL's). Also some comments were improved.


wrooot:
https://issues.igniterealtime.org/browse/SPARK-2045

PR #413, opened 2018-01-14
Added error windows trigger when user  typed with user user which doesn't exist [SPARK-1895]

Alexander198961:
https://issues.igniterealtime.org/browse/SPARK-1895
we  will trigger error window when sendStanza has been failed ( For example send message to user which doesn't exist)


wrooot:
Closing in favor of https://github.com/igniterealtime/Spark/pull/506#issuecomment-674395833

PR #415, opened 2018-02-05
fix space issue

Alexander198961:
changed whitespace with encoding value for enable search for Strings with spaces (https://issues.igniterealtime.org/browse/SPARK-2050)


guusdk:
Can you please explain why is this change needed? I'd expect Spark to be able to handle spaces.

SPARK-2048 is an issue that relates to case sensitivity, not spaces. Did you mean to refer to a different issue number?

wrooot:
My guess was that Spark can't search for last name as it doesn't recognize spaces. But lets wait for Alexander's reply. I haven't tested this change though.

wrooot:
I have just tried this change and it didn't resolve the issue.

Alexander198961:
I had some problem with search when I use whitespace for example "milan ac" is not working but when I type "milan%20ac" it returns row to me 
not sure but when I have created users it openfire server it all displayed to me with lower case:

wrooot:
I have a user Name Last (his username is namelast). If i search for name, it returns it. If i search for last, it doesn't. If i search for Last, it does. If i search for name last or Name last or name Last, it doesn't. But it returns result, if i search for Name Last. So it works with spaces in names (not in usernames), but it is case sensitive. If i create a user with a space in its username "user ok", it doesn't return result when i do user ok search. So there is a second issue that search can't search for usernames with spaces in it. Though i can workaround this by searching for user\20ok.
It seems you are trying to fix the second issue and not the one you have linked to. I can file a ticket for this second issue.

wrooot:
But your change breaks ability to search for Name Last. It doesn't return everything even when i put correct case.

wrooot:
Created SPARK-2050, fixed PR message. But the issues with searching for names with spaces should be fixed.

wrooot:
Alexander, maybe you should first test everything in your fork before committing every fix into this PR. Or at least report when you are done, so i would know when i can test your changes.

Alexander198961:
ok will do

Alexander198961:
is search option working  in the last Spark version ?

wrooot:
Do you mean 2.8.3? There were a few reports about problems with it. But it is either related to AD or MySQL. It works for me with the embedded database and no AD.

PR #418, opened 2018-02-27
Fileupload plugin: fix invalid port bug

dnsdhrj:
Fileupload plugin can upload file only once, and the next upload fails because of this bug.
`Putmethod.getURI()` seems to omit port number if the port of `Protocol` matches with that of the instance.
So I assumed that if `URI.getPort()` returns -1, the protocol need not be changed.

### DIFF
--- a/plugins/fileupload/src/main/java/org/jivesoftware/spark/plugin/fileupload/ChatRoomDecorator.java
+++ b/plugins/fileupload/src/main/java/org/jivesoftware/spark/plugin/fileupload/ChatRoomDecorator.java
@@ -184,9 +184,13 @@ public class ChatRoomDecorator
 
         try {
             PutMethod put = new PutMethod(response.putUrl);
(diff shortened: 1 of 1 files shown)


wrooot:
Thanks. I will let Guus review this and file a ticket in JIRA probably.

wrooot:
Wonder if @deleolajide new version of plugin already has a fix for this.

wrooot:
Dele, i have merged the patch. It fixes the bug with subsequent uploads and the other PR also fixed the compatibility issue. Not sure if you have anything else to do. But you can :) Thanks, dnsdhrj!

deleolajide:
That's it. We are good to go. That was exactly what I did. Thanks dnsdhrj!

wrooot:
How about another challenge? ;) https://issues.igniterealtime.org/browse/SPARK-2059

dnsdhrj:
Oh, I assume @deleolajide is working on it? Btw, it would be some time later I'll be able to fix it.

wrooot:
Sorry, forgot to mention that we already got suggestion how to fix it and Dele has approved.

PR #420, opened 2018-03-08
SPARK-1585: Add support for XEP-0280 'Message Carbons'

guusdk:


wrooot:
Carbons kind of work in Spark. But only if you already have a chat window opened with a person you should receive carbons from/to. If the chat window is not opened, either nothing happens or Spark opens chat window with that person, but new messages are not appearing there.

PR #425, opened 2018-04-21
[SPARK-2001] Add a popup for accepting certificate from a connection

Alameyo:


Alameyo:
I hope such solution looks good.

obraz

Alameyo:
Aaah, now I see that I removed implementation for Add chain button, but I will add it as separate PR.

Alameyo:
I don't think it's too much information for user in this dialog. Anyway it's something that casual user usually will need to go through this dialog only once and then maybe again after few years when certificate will expire. Unlike with web browsing where you often connect to different websites/servers, here you usually establish connection connection with one server which have your account. Some users might have few accounts on different servers but still I don't think it would be too much hassle for them.

wrooot:
How can casual user decide whether to press Add or Add chain?

guusdk:
I love the layout, looks good! I am with wroot on the complexity thing. Less is more.

Alameyo:
I like to keep user informed on what he is trusting. Anyway I will work on creating advanced tab or dialog but I will create separate PR with that. Also I am going to investigate how to disable that warning about certificate path validation fail.

speedy01:
I'm late to the party, but Id love to see the popup display like jitsi client does.
jitsi1
and after you hit "show certificate"
jitsi2

really all we would need is something like the 2nd screen. it has all the info a user needs, along with an option to continue anyway, or add the certificate/exemption and continue.

wrooot:
Well, me and Guus believe "less is more" and end user shouldn't be greeted with a wall of text. Also, Alameyo believes a user should select whether he/she wants to add just a cert of a whole chain, so  there are two buttons on the dialog now. Btw, we don't have a Continue anyway option for a one time only login. I guess that would be even more complex to implement.

Alameyo:
Idea with adding whole chain is because of Java's path validation API will fail if end entity certificate is in TrustStore but issuer isn't. Workarounds are adding that certificate to exceptions (which means any validation isn't done, even expiration), or to add issuer certificate.

wrooot:
So, is it safier to just add the whole chain every time?

speedy01:
the point of view I was trying to get across  would be to present all the relative info in a single box.  The box would have a few simple, easy to understand options.  In the case of the 2nd window; it gives the warning, shows the cert info, and presents simple options to continue. :)   I

 don't think importing the chain is a good option;  Id rather see the exemption based on the cert.  The reason for this, is you could inadvertently import a chain during a mitm attack.  if so, than any cert created from the CA used in the attack would be considered "safe".

wrooot:
I wonder what the general practice is, say in other clients or even browsers. I think only cert is imported and then a client/browser checks the chain on its own. But that's not always the case. I had an issue with mobile Skype for Business app not being able to check the trust of a domain, because that domain didn't have whole chain installed and the client/Android was not doing an additional download, what browsers did without a problem. Also, Alameyo says Java API may fail with only the certificate present. In the case of a MITM attack you would still be presented with a new cert. So i think it would still alert about it (not sure how it works).

Alameyo:
So by default import only end-entity certificate as exception and then under "advanced" tab present whole chain with option to import it (or just part of it)? 

About how I see MitM attack here: 
If we import certificates inband then MitM can happen anyway if we add only end-entity cert or CA. That's the place where we decide first time to trust something. Ideally certificates should be added in other way, out of band so this dialog wasn't in my original project. Additional problem with accepting CA this way, would create possibility for attacker to forge certificates for different servers.

And about state after accepting only end-entity cert in TrustStore (though I might be wrong somewhere here):
If I am keeping only end entity certificate in TrustStore and if server will present his certificates chain, I can only compare end-entity cert with the one in TrustStore. It's easy to "steal" certificate from connection. In theory Mallory still lack of private key paired with public key in certificate so it can't use it for encryption and authentication of . Here however comes TLS, there are some  cipher suits which allow for unauthenticated exchange (though disabled by default in Java). There can also be some problems with revocation mechanisms.

wrooot:
``So by default import only end-entity certificate as exception and then under "advanced" tab present whole chain with option to import it (or just part of it)?

Yes. I think this is  the best option.

PR #428, opened 2018-04-21
Spark-2001 Suppress message with certificate path validation error

Alameyo:


wrooot:
The exception is still happening for me. I have updated the list of existing issues in the new comment here https://issues.igniterealtime.org/browse/SPARK-2001

wrooot:
I still get a visual popup behind the new certificate add dialog.

wrooot:
spark_certpath

wrooot:
I'm on Windows 7 x64. Trying to login to the Openfire server with its own generated self-signed certificates.

PR #432, opened 2018-08-24
SPARK-2001 Advanced button in cert popup

Alameyo:


Alameyo:
Oh I still didn't added autologin after adding certificates. While it is possible I feel that I have made otherwise nicely structured code a bit messy so I didn't included that changes. Maybe later I will come with some better solution there.

wrooot:
I have added my comments at https://issues.igniterealtime.org/browse/SPARK-2001

wrooot:
Filed autologin as https://issues.igniterealtime.org/browse/SPARK-2070

PR #439, opened 2018-09-28
SPARK-1991 By default self signed certificate is added to IdentityStore after creation

Alameyo:
Also I fixed bug that I introduced with previous PRs that you couldn't upload certificates in MutualAuth tab ( and that mean I should start writing tests at some point :) ).

Other than that some clean code changes and I tried to enforce "KeyStore" instead of "Keystore" names for consistency, but that maybe should go to separate PR?


Alameyo:
Testing wise, it adds checkbox. By default certs are added to IdentityStore but if checkbox will be selected it will save them to file.
save_cert_to_file

PR #441, opened 2018-12-05
SPARK-1990 Remove pki tab

Alameyo:
I resubmit this PR but I am not sure about merging it yet. It removes PKI tab, but previously there was objection that it can remove also SASL external. I am currently not sure whether SASL external works anyway for Spark (I tried with 2.8.3 and current build without success) but it might be also related to  [Openfire](https://discourse.igniterealtime.org/t/sasl-external-for-c2s-doesnt-work/79539) and I didn't tried with other servers.


guusdk:
Fixed in #491

PR #444, opened 2018-12-11
enable Emoticons SPARK-2068

Alexander198961:
How I tests:
On MAC (I tested only in the mac os )
1 mvn install on git clone
2 ./distribution/target/distribution-base/bin/startup.sh

Then check that emoticon is working


wrooot:
Maybe it now works on macOS, but it doesn't on Windows still. Thanks anyway ;)

PR #448, opened 2018-12-26
fix for SPARK-2043

Alexander198961:
I have tested one symbol and couple symbols send and seems working for me with chat room


Alexander198961:
https://issues.igniterealtime.org/browse/SPARK-2043

wrooot:
Thanks

PR #465, opened 2019-04-02
SPARK-2109 fix current history file name

wrooot:


wrooot:
Maybe this is not a complex change to review. It works for me (now file is being generated with user@server_current.xml name instead of broken userserver_current.xml, in the same scheme like the main one user@server.xml). But i'm not sure if i haven't broken anything with my brute fix :)

core/src/main/java/org/jivesoftware/sparkimpl/plugin/transcripts/ChatTranscripts.java
```diff
@@ -232,11 +232,8 @@ public static File getTranscriptFile(Jid jid) {
      * @param jid the jid of the user.
      * @return the current transcript file.
      */
-    public static File getCurrentHistoryFile(CharSequence jid) {
-        // Unescape Node
-        jid = UserManager.unescapeJID(jid);
-
-        return new File(SparkManager.getUserDirectory(), "transcripts/" + jid + "_current.xml");
+    public static File getCurrentHistoryFile(Jid jid) {
+        return new File(SparkManager.getUserDirectory(), "transcripts/" + jid.asUnescapedString() + "_current.xml");
```
Flowdalic: Whenever a JID appears as part of a filesystem path component, as here, you usually want to use `Jid.asUrlEncodedString()`. However I am not sure if this method is already available in the jxmpp version we use. If it is not available, then we may want to put a comment reminding us to potentially switch to this method once it becomes available.

wrooot: I will add comments

wrooot:
If i just to replace it with Jid.asUrlEncodedString() it can't resolve it. So, i guess this is not available yet? Or maybe i should add another import? All i can find about it is this commit https://github.com/igniterealtime/jxmpp/commit/10dd7bde6cd0b162c03d9b9254f4ce3359722d25

wrooot:
As there were no other suggestions or objection, i have added comments and will merge this.

PR #471, opened 2019-05-16
SPARK-2065: Show error when preferences cannot be saved.

guusdk:
At times, a plugin that integrates with the preferences panel fails to store its configurations. This has been observed with 'old' plugins, that suffer from class-not-found exceptions. When this occurs, Spark should notify the user, but otherwise ignore the issue.


guusdk:
Note that this fixes only one aspect of SPARK-2065, but not all of it.

PR #477, opened 2019-05-25
Add OWASP dependency check to build

adiaholic:
The check is currenty optional and placed
inside the profile 'owasp-check'.
One can directly invoke the plugin via
 `mvn org.owasp:dependency-check-maven:check`
An attempt at resolving SPARK-2113.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -263,4 +263,29 @@
         </repository>
     </repositories>
 
+    <profiles>
+        <profile>
+        <id>owasp-check</id>
+        <build>
+            <plugins>
+	            <plugin>
+    	            <groupId>org.owasp</groupId>
+        	        <artifactId>dependency-check-maven</artifactId>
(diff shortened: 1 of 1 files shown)


guusdk:
I'm assuming that by adding a profile like you did, the intended usage is now: 

    mvn clean package -Powasp-check

Would it make sense to add this to our Travis build?

adiaholic:
Yes, you are right. The intended usage is 
`mvn clean package -Powasp-check`
But this will suffice our needs only on our local machines where we manually run it through terminal,
whereas travis might not be able to invoke this profile because it is not set as default.
So alternatively what we can do is invoke this profile by appending
`before_install:
  -  mvn clean package -Powasp-check`
to travis.yml.
As I am not much familiar with Travis, there might be a different and better solution altogether.
Let me know if this is proper way :)

PR #481, opened 2020-03-19
OpenJDK-13.0/NetBeans-11.3 support

g-stefan:
Modifications/update to be able to compile/use runtime of OpenJDK-13.0 
I used NetBeans  11.3 for IDE (maven support)
Some dependency/modules update


wrooot:
@guusdk what do you think?

PR #482, opened 2020-04-11
SPARK-2146: Bouncy Castle 1.65 instead of 1.57

Neustradamus:
Bouncy Castle 1.65 instead of 1.57: https://www.bouncycastle.org/releasenotes.html
- http://www.bouncycastle.org/latest_releases.html

CVEs: https://www.cvedetails.com/vulnerability-list/vendor_id-7637/Bouncycastle.html

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -14,6 +14,10 @@
     <name>Spark Core</name>
     <description>The essentials of Spark, including the user interface and startup scripts, excluding most optional components</description>
 
+    <properties>
+        <bouncycastle.version>1.65</bouncycastle.version>
+    </properties>
+
     <build>
         <finalName>spark</finalName>
         <resources>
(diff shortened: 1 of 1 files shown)


wrooot:
@guusdk what do you think? Is it safe to just swap versions?

guusdk:
Yes, should not be a problem. Just like with JREs, we should generally upgrade to the latest release prior to releasing Spark.

wrooot:
Created https://issues.igniterealtime.org/browse/SPARK-2146

Neustradamus:
@wrooot: Thanks for this merging and the 1.66 update in the same time:
- https://github.com/igniterealtime/Spark/commit/6b89d28ef7afc9aab5f1d489e1f9ac6e0d363177

PR #488, opened 2020-06-09
Properties fixed

LaurentBarbareau:
Single quotes doubled when necessary (as mentioned into the header comment). This fix parameterized labels not filled properly.
Spelling.


wrooot:
Thanks

PR #491, opened 2020-08-02
SPARK-1990: Remove PKI tab from login settings

guusdk:
This tab has been replaced by Pavel with two other tabs.


wrooot:
What about https://github.com/igniterealtime/Spark/pull/441 though?

guusdk:
Ah, I simply didn't see that. This PR is basically a duplicate of that.

If SASL EXTERNAL is now broken, we'll have to fix that, but lets tackle that as a separate issue (if that indeed turns out to be the case).

PR #492, opened 2020-08-02
SPARK-2068: Restore emoticons

guusdk:
This commit fixes the distribution packaging to include the emoticons. Additionally, the code that verifies if emoticons exist in the user home directory is a bit more error resistant now.


guusdk:
@wrooot would you mind testing this? I'm not 100% sure if this fixes all occasions of the problem - but it at least fixes several of them. :)

wrooot:
Emoticons are back :)

PR #493, opened 2020-08-02
SPARK-1979: Restore distribution archive directory for 'sounds'

guusdk:
The 'sounds' directory (and some other files) used to live outside of the spark.jar file in Spark 2.8.3. This commit mostly restores this.

I'm leaving out the emoticons, as I think that has been tackled in SPARK-2068.

### DIFF
--- a/core/src/assembly/installation-directory.xml
+++ b/core/src/assembly/installation-directory.xml
@@ -40,6 +40,10 @@
             </excludes>
         </fileSet>
 
+        <fileSet>
+            <directory>${project.basedir}/src/main/resources/sounds/</directory>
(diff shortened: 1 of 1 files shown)


guusdk:
@wrooot With this fix, I'm getting sounds again in Spark. I'm not sure if this is a 'complete' fix. Can you try on your end too, please?

guusdk:
I've added one more issue to this PR, as it's largely the same issue: put back stuff in the distribution that used to be in there before we mavenized things.

wrooot:
SPARK-1978 about missing documentation added to this PR.

PR #494, opened 2020-08-03
SPARK-2070: Login automatically after accepting server cert

guusdk:
When an authentication attempt fails because of the server presenting a certificate that's not recognized, the end user gets asked to add the certificate.

When the end user chooses to accept the certificate, another authentication attempt should be performed automatically. Up until this commit, that did not occur.

To be able to act based on the choice of the user, the mechanism in which the dialog is shown had to be replaced. This removed some of the more elaborate functionality added by Pawel (that allowed for each intermediate certificate in the chain to be added individually). I've opted to simply add every certificate in the chain.


wrooot:
Works great!

PR #502, opened 2020-08-05
Disable Smacks support for Stream Management (for now)

guusdk:
We ideally enable Stream Management (see SPARK-2140), but for that, the reconnection logic in Spark probably needs work. I'm disabling it for now, until we have had a proper look.


guusdk:
Please do _not_ close SPARK-2140 after merging this PR. SPARK-2140 was created to make sure that we eventually revert this PR!

PR #503, opened 2020-08-05
SPARK-2141: Contacts should always be shown, even if they're not in a group

guusdk:
This commit restores functionality that already existed, but was ... disabled?

When a user is not in any group on the contact list, it is now displayed in a psuedo-group, called 'unfiled' (this is similar to the 'offline contacts' psuedo group).


guusdk:
@speedy01 tells me this problem also exists in 2.8.3

wrooot:
Actually, it was disabled because it would sometimes appear without any users in it https://issues.igniterealtime.org/browse/SPARK-1758

PR #506, opened 2020-08-15
SPARK-1895: Show errors in chat transcript.

guusdk:
When a mesasge including an error is received, display an error message to the end-user, so that they know something went wrong.


wrooot:
Should we close https://github.com/igniterealtime/Spark/pull/413 ?

guusdk:
Yeah, I think so. That has not progressed much anyway.

PR #509, opened 2020-08-17
Denote Spark 2.9.0 Release

akrherz:


akrherz:
The next doubt here is if bamboo can build all the release artifacts yet, but shall cross that bridge soon!

wrooot:
I have just marked https://issues.igniterealtime.org/browse/SPARK-1895 as Fixed. It should be included in the changelog. Also, would be great to fix https://issues.igniterealtime.org/browse/SPARK-2143 before releasing (Ilya in the forums says this is critical for them as his users reconnect often throughout a day).

The note at the top of changelog can be changed as Guus reworked the certificates part:
This version is not accepting self-signed certificates by default. If you are using such, your users will have to confirm to use it on a first login. Also OTR and Spellchecker plugins are not included anymore (have no current Maven artifacts to use).

akrherz:
Added SPARK-1895 to the changelog and dropped the certificate note.

PR #523, opened 2020-09-02
Add files via upload

RJRasmussen1:
Change Spark-parent POM.xml to include maven-assembly-plugin, so that a single jar with all dependencies can be generated via mvn package. jar can be executed with java -jar spark-with-dependencies.jar

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
 
     <groupId>org.igniterealtime.spark</groupId>
     <artifactId>parent</artifactId>
-    <version>2.9.3-SNAPSHOT</version>
+    <version>2.9.0</version>
     <packaging>pom</packaging>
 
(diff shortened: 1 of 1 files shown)


guusdk:
I'm not sure about this: I assume that there's more to having a properly working Spark instance than just having a runnable JAR file - that's why the 'distribution' module is there. My fear is that by merging this PR, we'd get some kind of half-product: a jar that _is_ executable, but that's missing functionality. I dread needing to support that.

I have no issue with having an executable jar - but maybe there's more to it? If so, we should probably move it to the 'distribution' module?

Also: the changes to the native2ascii appear unrelated at first glance.

RJRasmussen1:
@guusdk Im working on testing it today and tomorrow. Will let you know if there is any missing functionality

PR #530, opened 2020-09-27
SPARK-2169: Rendering messages in chat area

R87A:
Optimize messages rendering in chat area.
Make it possible to correctly handle network addresses containing spaces when enclosed in quotes.

### DIFF
--- a/core/src/main/java/org/jivesoftware/spark/ui/MessageEntry.java
+++ b/core/src/main/java/org/jivesoftware/spark/ui/MessageEntry.java
@@ -159,70 +159,61 @@ public class MessageEntry extends TimeStampedEntry
                 // for now, we'll not visually differentiate between quotes and non-quotes.
                 for ( final String line : block.lines)
                 {
-                    int i = 0;
+                    int to = 0;
                     do
                     {
-                        // Skip through prepending whitespace.
-                        if ( Character.isWhitespace( line.charAt( i ) ) )
-                        {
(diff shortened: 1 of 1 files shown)


wrooot:
Thanks again. I have noticed a small side effect https://issues.igniterealtime.org/browse/SPARK-2170

Btw, do you join our group chat at http://www.igniterealtime.org/support/group_chat.jsp ? We can create a user for you and then you can use any XMPP client to join, if you don't want to use the anonymous web browser version. @akrherz can arrange you an account there.

Or do you prefer some other way of communication? There are groups in GitHub, and i tried to create one in the past, but it never took off. I have invited you into Spark Developers team here. If you accept if, you can go to https://github.com/orgs/igniterealtime/teams/spark-developers/discussions and we can discuss anything there. Maybe you can tell about what other issues you are planning to work on, etc. I was going to ask if you would be willing to take a look at a few major bugs since 2.9.0. fixing one or few of them would warrant the release of 2.9.3.

https://issues.igniterealtime.org/browse/SPARK-2143
https://issues.igniterealtime.org/browse/SPARK-2153
https://issues.igniterealtime.org/browse/SPARK-2162

Anyway, thanks for working on Spark. Any even small fix or just a cleanup is appreciated :+1:

wrooot:
Of course, our https://discourse.igniterealtime.org forums is also a place to discuss things. Especially with users.

PR #531, opened 2020-09-28
SPARK-2170: Fix converting emote expressions to icons

R87A:
It was a minor algorithmic error.


wrooot:
Thanks, this fixed it.

PR #532, opened 2020-09-28
SPARK-2162: Fix history issues for contacts with spaces in usernames

R87A:
Use unescaped JID to prevent creating files with backslashes in the name.


wrooot:
Fixed. Thanks a lot, again! :+1:

PR #533, opened 2020-10-01
SPARK-2153: Fix private messaging in group chat.

R87A:
To address private rooms in a group chat, you need to use full JID instead of bare JID.
[See section 7.5 of XEP-0045: Multi-User Chat](https://xmpp.org/extensions/xep-0045.html#privatemessage)


wrooot:
Thanks again :100:

PR #535, opened 2020-10-05
SPARK-2143: Fix contacts duplication after reconnect

R87A:
I hope one active forum user will be happy. [SPARK-2143](https://issues.igniterealtime.org/browse/SPARK-2143)

### DIFF
--- a/core/src/main/java/org/jivesoftware/spark/ui/ContactGroup.java
+++ b/core/src/main/java/org/jivesoftware/spark/ui/ContactGroup.java
@@ -1,12 +1,12 @@
 /**
  * Copyright (C) 2004-2011 Jive Software. All rights reserved.
- *
+ * <p>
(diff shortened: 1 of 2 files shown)


wrooot:
Thanks. I wonder is it possible for contacts to turn not red after a few seconds during reconnecting. Now they stay red until you reconnect. Not a huge problem. After a reconnect all now offline contacts are not red, so this is good. I have asked Ilya to test new build. The only side effect i have noticed, that it doesn't switch back to Online after reconnect and is stuck in Away status and you have to switch it to Online manually.

wrooot:
Ilya also provided some feedback. Side effect with duplicating tabs if you try sending messages to a tab that was open before reconnection, etc. https://discourse.igniterealtime.org/t/spark-2-9-0-after-reconnecting-some-users-change-their-status-to-red/88485/12

PR #550, opened 2020-10-31
Fix certchain validation

guusdk:


guusdk:
With this state, I'm still seeing issue SPARK-2187 which I'd like to fix before merging.

Also, with the improved checks, I'm now seeing validation failures caused by a certificate not providing OCSP data. When Spark is configured to verify OCSP, this now causes the validation to fail. I'm not sure if this is correct. Does that setting in Spark mean: "check if OCSP data provided by cert" or "Check OCSP, fail if no data provided" ?

wrooot:
@Alameyo do you remember how it was setup? The OCSP check.

guusdk:
It appears that the OCSP check failed because of a bug in our code (SPARK-2188). I've now fixed that.

I think this PR is now ready for a merge. There are still some minor issues, but what is in this PR should be a good improvement over what we had.

@wrooot I'd suggest to quickly build a version with this, and ask the usual suspects to do a round of testing with regards to different TLS scenarios.

Alameyo:
Seems alright.

PR #556, opened 2020-11-05
SPARK-2005: Query ClientControl for additions to plugin blacklist

guusdk:
This commit augments the blacklist of plugins with the response to a query to the clientControl plugin for Openfire.

This functionality is added to the clientControl plugin in what will probably be release 2.1.6. See https://github.com/igniterealtime/openfire-clientControl-plugin/pull/18


wrooot (approved):
As i understand this creates a framework in Spark to blacklist any plugin that will be added as an option to Client Control plugin without any further modifications to Spark itself?

PR #558, opened 2020-11-06
Auto accept file transfer

guusdk:


guusdk:
I'm going to merge this without @wrooot's approval, as I need a nightly that contains this feature with some haste.

PR #565, opened 2020-11-14
update Jira links to https + atlassian.net address

akrherz:


akrherz:
As with igniterealtime/openfire#1746, I am OK with this PR being declined.

PR #577, opened 2020-11-25
Run the electron download in a thread

deleolajide:
replace sun.misc.BASE64Decoder with javax.xml.bind.DatatypeConverter


wrooot:
This bundles Pade Meetings plugin again and also attempts to avoid UI locking up by downloading Electron files in a separate thread.

PR #594, opened 2021-01-07
Updated Russian translation for Spark 2.9.5

Plyha:
More than 5 years have passed since the last update of the translation. A lot of the text was in English. I took the spark_i18n.properties file as a basis and added the existing translation as well as translated new phrases into Russian


wrooot:
Thanks

PR #636, opened 2021-06-23
SPARK-2217 Update Russian translation

Plyha:
Also added translation for Spark plugins.
And it seems there was a spelling mistake in some Fastpath languages ​​(Type)


wrooot:
I don't think this is a typo. You need to use different letters on the same screen for Alt-* commands to work. If Title and Type both use T for that (which &Title means), then you won't be able to use Alt-T to reach Type field.

Plyha:
What do you think might cause some text to stop displaying correctly?
image

wrooot:
Encoding maybe. Maybe this line was copied from some place that used different encoding that usual. Try deleting it and type it manually.

Plyha:
The English and German characters look readable.But the Cyrillic alphabet is not displayed. I think this is due to the fact that this text becomes bold.

wrooot:
This might be some limitation of Java 8 as we are currently using a workaround to make non-latin translations work ok with Java 8.

PR #637, opened 2021-06-25
Replace icon on-phone

Plyha:
I guess I created the "on-phone" icon incorrectly, the design is not consistent. Wroot, What do you think about this icon?
Before
https://yadi.sk/i/wff1yAm1E3IObQ
After
https://yadi.sk/i/vMecQodd-KP9dg


Plyha:
Or just a phone?
image

PR #641, opened 2021-07-02
Fix ignoring of line breaks

oLucasPopov:
When a message had more than one line break, like this:

```
a

b

c
```
the following problem used to happen:

before_2

now that problem has been fixed, it will appear like this, as it should:
image

explanation of what I did, if necessary:
first, I changed **new StringTokenizer(text, "\n", false);** to **new StringTokenizer(text, "\n", true);** for it to return the delimiters as tokens, and it caused a problem of doubling each line-breaks.

For that reason, I refactored to add a new line only when finished the insertion on the lines of the block.


Plyha:
I tested with Spark 2.9.4 and Spark 2.8.3.
Thanks for your contribution to Spark!

PR #652, opened 2021-12-10
SPARK-2237 Revive the 'spell checker' plugin (using Jitsi-provided dependencies)

guusdk:


Plyha:
Hi, thanks for the fix. A quick test showed that the plugin is now working. I also changed the encoding(
now non-english letters are displayed correctly) and updated the icon for Spark 3.0.0

PR #661, opened 2021-12-20
SPARK-2241: Allow SoundPreferences class to be deserialized

guusdk:
The XStream library that takes care of deserializing XML data to a class instance needs to be configured to allow certain classes to be instantiated.


Plyha:
I am unfortunately unable to reproduce this. I've tried with both the new library(1.4.18) and the old(1.4.16). I am not getting this error in the IDE and in the logs.
Perhaps this is only relevant for Linux?

guusdk:
I don't know, to be honest. Might be that I've got different settings enabled in that 'sound' section?

Plyha:
And after your fix the bug went away? Can I merge your commit?

guusdk:
Yes, that stacktrace went away. I have various other issues, but I think
that they are unrelated.

PR #664, opened 2021-12-23
Bump install4j runtime lib to 9.0.5

Plyha:


guusdk:
We should test if this is compatible with the distributables of Spark that are generated by Install4j.

Plyha:
Unfortunately, I do not have a "bamboo" account to find out which version of install4j we are using.
I can assume that we are currently using version 8.0.7 - https://igniterealtime.atlassian.net/browse/SPARK-2205
It would be helpful but not critical to update install4j to the latest version due to Apple M1 support and the new FlatLaf theme (which is used in Spark)

Plyha:
@akrherz  Daryl, maybe you can help with the install4j update? I would happily test this.

akrherz:
@Plyha our default bamboo install4j version is 9.0.5, I am unsure though how to test PRs within bamboo.

Plyha:
Yes, I see Spark is installed with install4j 9.0.5.
Hmm, I wonder what is the effect of the install4j version in the pom.xml then?

`[INFO] Logger started at 2021-12-24 22:53:25
[INFO] Executable name: C:\Users\Plyha\Downloads\spark_3_0_0-20211218-with-jre.exe
[INFO] install4j version: 9.0.5 (build 9133)`

Plyha:
My bad.
I thought that the install4j version and the version in the pom.xml should be the same. in pom.xml install4j Runtime version. I close this commit.
https://mvnrepository.com/artifact/com.install4j/install4j-runtime?repo=ej-technologies

PR #666, opened 2021-12-27
Boolean form field fixes

Flowdalic:


Plyha:
Thanks for releasing Smack 4.4.5
Removed logic

PR #674, opened 2022-01-21
SPARK-2256 Add control over idle settings

Plyha:
In my company, some users turn off idle time or make the time very long. It would be helpful for administrators to manage these settings.


Plyha:
https://github.com/igniterealtime/openfire-clientControl-plugin/issues/31

PR #689, opened 2022-07-09
SPARK-2272 SPARK-2270 Remove unused methods in URLFileSystem.class

Plyha:
These methods are not used and use reflection.
This raises a warning for Java 11-15. Due to the strong encapsulation introduced starting from jdk 16 some non-public classes, methods and fields of the java.* package are no longer available via reflection.


guusdk:
Thanks!

PR #692, opened 2022-07-18
SPARK-2266 Always accepts self-signed certificates

Plyha:
I think with a self-signed certificate, an exception should be thrown in order to get into the catch block.


Plyha:
So, a quick test showed that self-signed certificates are now checked and it seems to work.

https://user-images.githubusercontent.com/71222850/192148887-32643a97-08f0-4b00-aaf5-3a34bbcd2f6e.mp4

PR #695, opened 2022-07-19
SPARK-2275 Replace deprecated custom ulit class Base64

Plyha:
Since Java 8 there is Base64 util class https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html
I think we should use it.


guusdk (approved):

Plyha:
Thanks!

PR #715, opened 2022-09-01
SPARK-2098 SPARK-2144 Add a button to restart Spark if network error happens

Plyha:
@wrooot @guusdk  Hello, how do you like this idea?
I don't think it's worth making a separate button.
bandicam 2022-09-01 21-23-33-145 (1)


Plyha:
Added buttons for other styles. I think in the future it is necessary to leave only one style of reconnection)

PR #739, opened 2022-10-06
SPARK-2276 SPARK-2016 Install4j 64-bit Setup doesn't uninstall previous 32-bit Version

Plyha:
Spark 64bit should uninstall past versions of Spark 32bit and vice versa

https://stackoverflow.com/questions/57557144/install4j-64-bit-setup-doesnt-uninstall-previous-32-bit-version


akrherz:
:eyes: Whoa, @guusdk , we likely need this for Openfire too!  Great job @Plyha !

Plyha:
Thanks!

PR #748, opened 2022-10-18
Add Preferences Panel in SparkMeet

deleolajide:
Adding ability to modify Jitsi Meet URL from Spark preferences

### DIFF
--- a/plugins/meet/pom.xml
+++ b/plugins/meet/pom.xml
@@ -10,7 +10,7 @@
     </parent>
 
     <artifactId>meet</artifactId>
-    <version>0.0.8</version>
+    <version>0.0.9</version>
 
     <name>Pade Meetings Plugin</name>
     <description>Adds support for Pade Meetings to the Spark IM client.</description>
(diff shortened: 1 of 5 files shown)


Plyha:
I installed Pade on my Openfire server, I'm trying to start a video conference, but after logging in, I immediately get an error and reconnect. I read the manual, users added to the room.Can you suggest what I'm doing wrong?

Also, after each connection to the video conference, all users who have an active dialogue with me receive a notification about my disconnection




image


image

deleolajide:
Sorry, documentation need more work. Enable debug in openfire logging and look at log file.

That is not good.  
image

You need
image

You should have client sessions for  **jvb** and **focus** users. If you are using LDAP, you have to create the **jvb** and **focus** users manually
image

Both users should join a groupchat called **ofmeet**
image

Allow non-tls client connections
image


You don't need to install pade to use SparkMeet. You can use the public Jitsi Meet service. Use https://meet.jit.si/ as your URL

image

deleolajide:
Thank you for the review and suggestions. 

However, I  want to remain compatible with existing installed versions and read the property file already created in HOME/Spark

There is a long list of parameters that can be used
https://github.com/jitsi/jitsi-meet/blob/master/config.js
https://github.com/jitsi/jitsi-meet/blob/master/interface_config.js

Maybe later in the future, I can create a proper config UI like I have in Pade client

image

PR #749, opened 2022-10-19
HttpUploadPlugin 0.4 Fix freeze UI. Upload each file in a separate thread.Add dialog

Plyha:
@deleolajide Do you mind if I make changes to your plugin?

Now when uploading a file on the server interface freezes and I can't do anything. It is also not clear what error occurred.
This fix:
-adds a window with an error
-uploads each file in a separate thread.


deleolajide:
>Do you mind if I make changes to your plugin?
Not at all. Please free free to improve it 👍

PR #817, opened 2023-03-14
SPARK-2326: Upgrade Apache HttpClient from 3.1 to 5.2.1

guusdk:
Beware! The API in HttpClient has changed significantly. I did not fully test these changes.


Plyha:
A quick test showed that Spark was also unable to find updates. I did some analytics on this problem, if you look, it will be great!
https://igniterealtime.atlassian.net/browse/SPARK-1851

PR #838, opened 2023-05-04
SPARK-2329: Support XMPP URI for joining group chat

guusdk:


Plyha:
It's great that you can now join a room via a URI. Do you mind if I add buttons so that you can quickly copy the link to the room?)





image

PR #841, opened 2023-05-29
Improve error message when failing to join MUC

guusdk:
When joining a MUC fails, an error message is shown. That message should include a reference to the server that is causing the error.


stokito:
Already fixed in 24c4fa5ff1dd9e907e90fb9ec26f498247a730ee

PR #860, opened 2024-08-02
docbook/installation-debian.sgml: improve commands and ru translation

stokito:
Instead of the old Sun Java use the default-jre package. Add a link to GitHub Releases page from which to get the deb package. Add wget command to download.
Use apt to install the package: it will also install dependencies (none for now but for a future). Add Russian translation.

The .deb file can be improved and just declare own dependency to the JRE.


Plyha:
Thank you!
Added

PR #861, opened 2024-08-04
spark-core: separate message when account registration is not allowed

stokito:


Plyha:
Are you using Openfire?

I tried to disable the registration of account and I see condition=forbidden

image

Plyha:
I think we need to add both conditions because the server’s implementation may be different.

thanks for your PR.

PR #866, opened 2024-08-11
Fix Maven build

stokito:
Currently the build is broken (surprise):

```
The POM for org.minidns:minidns-core:jar:1.0.3-SNAPSHOT is missing, no dependency information available
```

You may not see the error because the dependency is in your maven local .m2 repository. So you can reproduce this be removing it from local cache: `rm -fr ~/.m2/repository/org/minidns/`.

The problem is complicated here is that the spark-core declares a dependency to the minidns as a range: `[1.0.0, 1.0.999]`
https://github.com/igniterealtime/Smack/blob/master/build.gradle#L149C21-L149C37

This is dangerous to use in libraries. So the Maven tries to find all versions of the dependency. And here comes the second problem: the pom.xml has two additional repos:

```
    <repositories>
        <repository>
            <id>maven-snapshots-repo</id>
            <url>https://oss.sonatype.org/content/repositories/snapshots</url>
        </repository>
        <repository>
            <id>ej-technologies</id>
            <url>https://maven.ej-technologies.com/repository/</url>
        </repository>
    </repositories>
```

The Maven goes into the maven-snapshots-repo and tries to resolve the 1.0.3-SNAPSHOT which is not exists there but it was referenced from some other dependency. In fact the Spark dosn't use snapshot versions (and it shouldn't) so the simplest solution is to just remove the maven-snapshots-repo as unused. It looks like this fixed the build.

Additionally it would be better to not use the range versions in the Smack but instead stick to the latest available today.

I also made the `mvn dependency:analyze-only` to see that some dependency is used in a compile time but it actually comes transitionally and once the intermediate dependency removes the dependency then the build become broken. So I added those dependencies explicitly to the pom.xml

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -239,7 +239,7 @@
         <dependency>
             <groupId>com.install4j</groupId>
             <artifactId>install4j-runtime</artifactId>
-            <version>10.0.4</version>
+            <version>10.0.8</version>
         </dependency>
 
         <dependency>
@@ -294,10 +294,6 @@
     </dependencies>
     
     <repositories>
-        <repository>
-            <id>maven-snapshots-repo</id>
-            <url>https://oss.sonatype.org/content/repositories/snapshots</url>
-        </repository>
         <repository>
             <id>ej-technologies</id>
             <url>https://maven.ej-technologies.com/repository/</url>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -132,6 +132,21 @@
             <version>${dependency.smack.version}</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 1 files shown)


akrherz:
Thanks @stokito , if I understand you correctly, you believe Smack needs updated as well?  If so, can you kindly submit a PR there?

akrherz:
@stokito I'm not really a developer on Spark, I just help out with the ignitrealtime organization

Plyha:
Hello!) Thanks for your PR, everything works, except install4j, for some reason I can't download it, error 403, maybe it's related to my location (Russia)

PR #869, opened 2024-08-11
ContactList hide error when trying to get last activity

stokito:


Plyha:
Can you please show what you corrected with some example?

PR #873, opened 2024-08-23
SPARK-2340 refactoring: GraphicUtils.scaleImageIcon(): scale only if needed

stokito:
Avoid allocation `new ImageIcon(img)` when the image can be used as is without scaling.
The scaleImageIcon is called on each contact popup so if we can optimize it a little bit that would be good.
The resulted image is not modified anywhere after so it's safe to return the original image.


Plyha:
I completely agree with you, thanks for your PR!

PR #875, opened 2024-08-24
SPARK-2338 Refactor: rename the confusing "Old SSL" to "Direct TLS"

stokito:
The code uses a term "Old SSL" for Direct TLS connections on port 5223.
This is supported by [XEP-0368](https://xmpp.org/extensions/xep-0368.html) and may be not necessary for old clients.
To make the code follow same terminology as XEP I renamed the LocalPreferences.isSSL() to isDirectTls(). The property name in properties file internally remains same.

Additionally extract the similar code to configure TLS in the Login and AccountCreationWizard.


Plyha:
As I understand it, Direct TLS is depricated, I don't think it will be added to Smack, but it's worth telling @Flowdalic  about it

PR #877, opened 2024-08-24
SPARK-2341 Improve signup wizzard

stokito:
First changes for the https://igniterealtime.atlassian.net/browse/SPARK-2341
I'll continue on this in a few months later because the current changes are already useful and will allow to users to get an account more easily so the main goal is achieved.


Plyha:
Excellent! Is it possible to make domains taken from existing accounts on the computer? For example, as an administrator in my company, I did not want users to create accounts on the Internet.

Plyha:
What you are doing is great! But this list is difficult to maintain. I'm afraid there will be people who want to add their domains to the list, and Spark's release may not happen in the near future. I think the list should be compiled based on existing accounts, or the administrator could add his domains to the spark configuration file.
image

Plyha:
I think the idea of ​​a computer domain + a DNS request to the controller is an excellent solution.

PR #879, opened 2024-08-26
SPARK-2343 Show error "Contact search service is not available"

stokito:
Partial fix for the https://igniterealtime.atlassian.net/browse/SPARK-2343


Plyha:
Great!
I would really like to hide the search field if XEP-0055 is not supported, but I don’t know where to put the secure/non-secure connection icon then.
image

Plyha:
If I turn off the search, the icon will be in the middle of the field. Perhaps it should be moved higher to the status...
image

PR #881, opened 2024-08-31
Use APP_NAME from spark.properties

stokito:
Use APP_NAME from spark.properties instead of APPLICATION_NAME from default.properties.
This is confusing to have two similar properties with the same value in two separate configs.


Plyha:
I agree with you, thanks for your PR

PR #884, opened 2024-09-08
SPARK-2346 make default-jre or any other JRE for Linux Deb

stokito:
https://igniterealtime.atlassian.net/browse/SPARK-2346

The launcher has a the std out/err log redirection that doesn’t work and when I starting the Spark I see the exception: 

 

    java.io.FileNotFoundException: log file directory '/opt/Spark/logs' doesn't exist.
    	at com.exe4j.runtime.LauncherEngine.checkRedirectionFile(LauncherEngine.java:196)
    	at com.exe4j.runtime.LauncherEngine.doRedirection(LauncherEngine.java:172)
    	at com.exe4j.runtime.LauncherEngine.launch(LauncherEngine.java:59)
    	at com.install4j.runtime.launcher.UnixLauncher.start(UnixLauncher.java:69)
    	at install4j.org.jivesoftware.launcher.Startup.main(Unknown Source)

### DIFF
--- a/distribution/src/installer/spark.install4j
+++ b/distribution/src/installer/spark.install4j
@@ -834,7 +834,7 @@ return console.askYesNo(message, true);
       </exclude>
       <jreBundle jreBundleSource="none" />
     </linuxRPM>
-    <linuxDeb name="Linux Deb Archive" id="166" dependencies="default-jre | java-runtime (>= 8)">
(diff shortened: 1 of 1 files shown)


Plyha:
so, i disable redirect stdout

image

and added your dependency "default-jre | java-runtime (>= 8)", but you can see that the config displays the symbol code instead of the symbol itself.

image|

In any case, if it doesn’t work, we’ll roll back the changes with the dependency.

PR #904, opened 2025-09-20
SPARK-2349: fix NPE when adding yourself to contact list

stokito:
In this case you'll get the presence from PresenceManager constants that doesn't have the from field:

    PresenceStanza id=7QUYG-13,type=available,mode=available,status=Online,prio=1,]

This make a two NPEs.


guusdk (changes requested):
This looks good (even if it's a bit of an edge case) but please create a JIRA issue (and change the commit message to reference the issue identifier) so that this change becomes part of the release notes.

PR #905, opened 2025-09-20
i18n: remove preamble from files

stokito:
The preamble is not really useful and it definitely must not be copied to translations. Instead I added a link to the Transifex so that users will see the proper way to submit translations.
This also reduced size of i18n from 976K down to 956K i.e. saved 20kb.


guusdk (approved):
Thanks!

PR #977, opened 2026-06-24
Security hardening, dependency updates and deprecation cleanup

imurd:
**This branch collects a round of maintenance work:**
TLS fix, dependency bumps, and cleanup of deprecated and noisy APIs, plus a few tests.

**The main change is on the security side.** 
AcceptAllCertsConnectionManager disabled all TLS certificate and hostname verification, and it was used for the update descriptor, the installer download, the plugin list, plugin downloads and inline image downloads. It is removed and those paths now use the default validating connection manager. The update URL is also switched to HTTPS.

**Dependencies are updated:** 
FlatLaf 3.1.1 to 3.7.1,
BouncyCastle 1.79 to 1.84, 
JNA 5.12.1 to 5.19.1, 
dom4j 2.1.4 to 2.2.0. 

CertificateModel is adapted to BouncyCastle 1.84 by replacing the deprecated AuthorityKeyIdentifier.getKeyIdentifier() with getKeyIdentifierOctets().

Compiler deprecation warnings are resolved: AWT modifier masks migrated to the extended DOWN variants and getMenuShortcutKeyMask to getMenuShortcutKeyMaskEx, JTextComponent.viewToModel replaced with viewToModel2D, Class.newInstance with getDeclaredConstructor().newInstance, and commons-lang3 endsWithAny with plain endsWith checks. The intentional ContactGroupListener.showPopup compatibility bridge is annotated with SuppressWarnings, and dead char-to-modifier-mask comparisons in JContactItemField are removed.

Console output in core and the active plugins is routed through the Log class instead of printStackTrace and System.out/err.println. The intentional pre-logger startup diagnostics in Spark are left as they are.

SwingX usage is dropped where standard APIs cover it: java.time is used for date arithmetic instead of SwingX DateUtils, JXList is switched to javax.swing.JList, and List is used in place of Vector where no API requires the latter.

Finally, tests are added for StringUtils.keyStroke2String, JContactItemField.validateChar/validateChars, and AvatarPanel image extension filtering.

PR #979, opened 2026-06-24
Bump FlatLaf, JNA and dom4j

imurd:
Another small, focused pull request split out from #977.

This bumps three dependencies in core:

FlatLaf from 3.1.1 to 3.7.1
JNA from 5.12.1 to 5.19.1 (both jna and jna-platform)
dom4j from 2.1.4 to 2.2.0

Only version numbers change, no source changes are needed for these.

BouncyCastle is deliberately left out of this PR, since dependabot already covers that bump to 1.84 in #967 and #968.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -193,7 +193,7 @@
         <dependency>
             <groupId>org.dom4j</groupId>
             <artifactId>dom4j</artifactId>
-            <version>2.1.4</version>
+            <version>2.2.0</version>
         </dependency>
         <dependency>
             <groupId>org.apache.httpcomponents.client5</groupId>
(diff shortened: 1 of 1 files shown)


stokito:
The next release will be 3.1 and I wanted to update the Flatlaf on the overnext release 3.2 along with Dark theme. The UI problems are hard to fix and users (especially old) may not like the look changes. So I want to make one release without significant UI changes and then another one with them.
The BouncyCastle I'm is in fact transitional dependency on the Smack XMPP library so I will update it only to more fresh patch version.

I'll pick the JNA update.

stokito:
d4b16eb7243c0c0551b8d632935376a78aac5eea

PR #1025, opened 2026-07-27
Fix contact JID handling and roster fallback

max199:
## Summary

- prevent `UserManager.escapeJID` from throwing for usernames without a domain
- retry contact creation with the regular roster subscription flow when subscription pre-approval is advertised but unavailable
- avoid showing a success dialog when contact creation fails
- add regression tests for bare and complete JIDs

## Problem

The add-contact dialog can silently do nothing when a server advertises roster subscription pre-approval, but Smack rejects `preApproveAndCreateEntry` with `FeatureNotSupportedException`. The exception was ignored and the regular `createItemAndRequestSubscription` flow was never attempted.

A username without a domain could also produce a null localpart in `escapeJID`, which was dereferenced before the caller appended the XMPP service domain.

## Testing

- `mvn clean verify`
- verified the resulting Spark distribution builds successfully

### DIFF
--- a/core/src/main/java/org/jivesoftware/spark/UserManager.java
+++ b/core/src/main/java/org/jivesoftware/spark/UserManager.java
@@ -280,6 +280,9 @@ public class UserManager {
         }
 
         String node = XmppStringUtils.parseLocalpart(jid);
+        if (node == null || node.isEmpty()) {
+            return jid;
+        }
         String restOfJID = jid.substring(node.length());
         String builder = XmppStringUtils.escapeLocalpart(node) + restOfJID;
         return builder;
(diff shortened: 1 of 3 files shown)


stokito (approved):

stokito:
Hi, thank you for the contribution. Did you fixed any problems that you faced or just wanted to improve the code? Because I can't understand why you made the changes. So please clarify:
1. The exception was ignored... because it will never ever thrown because we just checked if the feature is supported.
2. Why would you have a username without a domain?

It looks like the PR doesn't do anything worse, so I merged. But please clarify what problems you tried to solve.

PR #1026, opened 2026-07-27
Make Spark updates architecture-aware

max199:
## Summary

- report normalized runtime architecture values (`x86`, `x64`, `arm64`) to the Spark updater component on every supported operating system
- fix stable automatic updates being skipped when beta checking is disabled
- clear the global updater state on early exits and cancellation paths
- compare Spark versions numerically, including pre-release qualifiers
- verify downloaded packages using the server-provided filename and SHA-256 before launch
- preserve compatibility with servers that do not yet provide the new metadata
- restore the official install4j Windows x86 media set and compatible x86 native libraries
- prevent the unfiltered resource copy from overwriting the resolved Spark version

## Architecture support

The updater protocol is architecture-aware for Windows, macOS and Linux. The client emits one of these normalized values when applicable:

    x86
    x64
    arm64

Package selection remains server-side. The companion Client Control pull request implements `spark.<os>.<arch>.client` properties with fallback to the existing generic properties:

https://github.com/igniterealtime/openfire-clientControl-plugin/pull/57

## Compatibility

Windows packages continue to use the standard install4j EXE flow. This pull request does not add a custom launcher, ZIP updater, PowerShell replacement mechanism or alternative restart path.

Older Client Control versions remain usable: if `fileName` and `sha256` are absent, Spark falls back to the existing download behavior.

## Testing

- `mvn clean verify`
- architecture normalization tests for i686/x86, amd64/x64 and aarch64/arm64
- stable, pre-release and numeric version comparison tests
- download filename and SHA-256 validation tests
- install4j 12.0.2 Windows x86 media build (`media id 186`)
- unattended installation of the generated x86 EXE
- verified `Spark.exe`, `starter.exe`, bundled Java and all included native libraries are PE x86
- loaded all 11 included x86 native libraries with the bundled Java runtime
- verified the stock `Spark.exe` launcher and `starter.exe -> Restarter -> Spark.exe` flow


core/src/main/java/org/jivesoftware/sparkimpl/updater/SparkVersion.java
```diff
@@ -108,9 +127,24 @@ else if (Spark.isMac()) {
         else {
             buf.append("<os>linux</os>");
         }
+        buf.element("arch", currentArchitecture());
         return buf;
     }
 
+    static String currentArchitecture() {
+        final String arch = System.getProperty("os.arch", "unknown").toLowerCase(Locale.ROOT);
+        if (arch.equals("x86") || arch.matches("i[3-6]86")) {
```
stokito, line 136: seriously, i386?

core/src/main/java/org/jivesoftware/sparkimpl/updater/SparkVersion.java
```diff
@@ -36,6 +37,8 @@ public class SparkVersion extends IQ {
     private String downloadURL;
     private String displayMessage;
     private String changeLogURL;
+    private String fileName;
+    private String sha256;
```
stokito, line 41: why?

stokito (changes requested):

stokito:
Mr. @Shepard199 are you a spammer? This changes are meaningless.

guusdk:
Thanks for taking the time to contribute to Spark, especially as a first-time contributor. We appreciate the effort that went into this.

After reviewing the PR, we've decided not to merge it in its current form, and will be closing it.

There are a few reasons for this:

* This PR bundles a number of different changes. Each of these deserves to be reviewed on its own. Smaller, focused pull requests are much easier to review, discuss, and merge.
* The PR adds and updates Windows DLLs. As a project, we have to be very careful about accepting binary artifacts, especially from contributors that we do not yet know. Without a fully reproducible process that allows us to independently generate and verify these binaries, we cannot safely accept them.
* The PR description appears to have been generated largely by AI. While AI can be useful when writing code or documentation, we prefer pull request descriptions that are concise. The current description contains a lot of detail, but it makes it difficult to identify the actual motivation, scope, and rationale for the changes that reviewers need to evaluate.

If you're interested in contributing further, we'd be happy to review smaller, self-contained pull requests. For example, separating out improvements to version comparison, update validation, or architecture detection into individual PRs would make review much more manageable. For changes involving packaged binaries, please first discuss the approach in an issue so we can agree on a process before code is submitted.

Thanks again for your interest in improving Spark, and we hope you'll consider contributing again in a more incremental way.

PR #1028, opened 2026-07-31
SPARK-2118: Move 'multiple resources' to 'view client version' 

guusdk:
This reverts the previous solution to show all resources for a user (which was displayed in the contactlist tooltip for each user - introducing confusing behavior in a very visible place).

Instead, the resources are now displayed in the 'view client version' menu item, which adds a descriptive disclaimer. This is a less visible place (which will lead to less confusion) and as a bonus, the functionality now no longer returns data from one client when multiple clients are online.

### DIFF
--- a/core/src/main/java/org/jivesoftware/sparkimpl/plugin/jabber/JabberVersion.java
+++ b/core/src/main/java/org/jivesoftware/sparkimpl/plugin/jabber/JabberVersion.java
@@ -1,5 +1,5 @@
-/*
- * Copyright (C) 2004-2011 Jive Software. All rights reserved.
+/**
+ * Copyright (C) 2004-2011 Jive Software, 2026 Ignite Realtime Foundation. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


guusdk:
To prevent merge conflicts, this PR also includes both commits from #1027.

core/src/main/java/org/jivesoftware/sparkimpl/plugin/jabber/JabberVersion.java
```diff
@@ -133,7 +134,15 @@ private void viewClient() {
...
             }
-            final Jid jid = presence.getFrom();
+
+            // Collect all available full JIDs for the selected user.
+            final Set<FullJid> allFullJids = SparkManager.getRoster()
+                .getAvailablePresences(presence.getFrom().asBareJid())
+                .stream()
+                .map(p -> p.getFrom().asFullJidIfPossible())
+                .filter(Objects::nonNull)
+                .collect(Collectors.toSet());
```
stokito, line 144: The presences are sorted by a prio but the set is is unordered.

stokito (approved):

stokito:
Merged, thank you

PR #1031, opened 2026-08-17
SPARK-2409: Bring the Asterisk-IM phone plugin back to life

mhterres:
Spark's phone plugin has been dormant since early 2026. This brings it back: it builds, it
packages as a real Spark plugin, and it talks to the Asterisk-IM Openfire plugin again.

## Background

The plugin's client layer used to live in an external jar (`org.jivesoftware.phone.client.*`).
In 7134422b1 those classes were replaced with in-repo stubs so the tree would compile —
`isPhoneEnabled()` became `return true`, both dial methods became empty, and
`RingEvent.getCallerIDName()` returned the literal `"TODO CallerIDName"`. Shortly after, in
48820b718, the module was commented out of the reactor altogether and has not been built
since.

So the plugin was in an odd state: enough UI to offer phone actions on every contact, and
nothing behind it. This fills the stubs in against Smack 4.5 and re-enables the module.

## Contents

Six commits, each independently reviewable:

1. **Package the phone plugin like every other Spark plugin** — the POM parented off the
   reactor root rather than `plugins/plugin`, so it declared `spark-core` itself and produced
   a plain jar with no plugin assembly, and it carried no `plugin.xml` at all. Even a built
   artifact could not have been loaded by the plugin manager.
2. **Parse the `phone-event` extension sent by Asterisk-IM** — the registered provider was an
   empty class and the three event objects were stubs.
3. **Send the `phone-action` IQ understood by Asterisk-IM** — there was a placeholder for the
   provider and nothing at all for the request.
4. **Implement `PhoneClient` against the Asterisk-IM plugin** — discover the phone component
   via disco, gate on the `http://jivesoftware.com/xmlns/phone` feature, dial, and dispatch
   incoming events to listeners.
5. **Restore the user's presence when a call ends** — the body of the hang-up handler was
   commented out, so Spark set the user "On the phone" and left them there for ever.
6. **Re-enable the phone plugin module.**

The module is re-enabled last, so no intermediate commit leaves the reactor broken.

## Verification

Built and run locally against the Asterisk-IM Openfire plugin, and the phone integration
worked.

## Notes for review

**The user-facing gate is deliberately strict.** `PhoneClient`'s constructor fails unless the
server publishes a `phone` component *and* the logged-in user is mapped to a device on the
plugin's Phone Mappings page. `PhonePlugin` already treats that failure as "no phone support"
and shows no phone UI, which seems better than the previous behaviour of offering phone
actions unconditionally.

**Dial requests must carry a `to`.** Without one they go to the user's own server and never
reach the plugin's IQ handler. Similarly, the server compares the `phone-action` type
attribute against the uppercase enum name, so `"DIAL"` is required — `"dial"` comes back as
`feature-not-implemented`.

**`PhoneEventExtension` declares `QNAME`.** Smack resolves an extension class to a QName
through that static field and only falls back to `ELEMENT` + `NAMESPACE`; without it,
`Stanza#getExtensions(Class)` fails at runtime with `NoSuchFieldException`.

**One piece is deliberately left out.** The old hang-up handler also reconciled the restored
presence against `UserIdlePlugin`'s desktop-lock and idle state. That code cannot be revived
as written: it calls `UserIdlePlugin.getDesktopLockStatus()`, which no longer exists anywhere
in the tree, and reads `UserIdlePlugin`'s private `latestPresence` and `pref` fields. Its
counterpart in `UserIdlePlugin#setOnline()` is also still commented out, carrying the marker
*"TODO commented out when PhonePlugin was not ported to Maven"* — the two halves were disabled
together as one unit. Reviving either means changing core, so I have kept it out of a PR whose
job is to make the plugin build and work again. Happy to follow up separately if you want that
behaviour back.

**Versioning follows the existing convention** — `minSparkVersion` 3.1.0 and `<java>11</java>`,
matching all 16 plugins that already ship a `plugin.xml`.

## Relationship to other work

The server side of this is igniterealtime/asterisk-im#21, which brings that plugin up to
Openfire 5 and Asterisk 16-23. The two are independent changes in separate repositories, but
they are meant to be used together.

### DIFF
--- a/plugins/phone/pom.xml
+++ b/plugins/phone/pom.xml
@@ -3,10 +3,10 @@
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
-        <groupId>org.igniterealtime.spark</groupId>
-        <artifactId>parent</artifactId>
+        <groupId>org.igniterealtime.spark.plugins</groupId>
+        <artifactId>plugin</artifactId>
         <version>3.1.0-SNAPSHOT</version>
-        <relativePath>../../pom.xml</relativePath>
+        <relativePath>../plugin/pom.xml</relativePath>
     </parent>
 
     <artifactId>phone</artifactId>
@@ -25,12 +25,7 @@
         </contributor>
     </contributors>
     <dependencies>
-        <dependency>
-            <groupId>org.igniterealtime.spark</groupId>
-            <artifactId>spark-core</artifactId>
-            <version>${project.parent.version}</version>
-            <scope>provided</scope>
-        </dependency>
+        <!-- spark-core is inherited from the plugin parent, in 'provided' scope. -->
         <dependency>
             <groupId>javax.media</groupId>
             <artifactId>jmf</artifactId>
@@ -38,4 +33,13 @@
             <scope>provided</scope>
         </dependency>
     </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+            </plugin>
+        </plugins>
+    </build>
 </project>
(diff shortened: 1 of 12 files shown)


stokito:
I squashed and merged, thank you!

